Issue #16, opened 2024-03-04
Auto detect CELLxGENE collection for source datasets


NoopDog:
Add a new field "CELLxGENE Collection" to source datasets. 

On the build catalog, call out to the CELLxGENE API for collections read them all in a try to match collections based on DOI. Save the collection id on the source dataset if it's DOI matches the source dataset DOI. 

Add a left hand filter, "in cellxgene" on source datasets with values "yes" / "no". Tag the source dataset with yes/no for this field depending on the presence or absence of the cellxgene collection is.

There are complexities here we can ignore for now like:

- DOI can change I think when a preprint gets published ( need to verify)

- it may be possible for multiple DOIs to point to the same paper (again need to verify)

NoopDog:
Complete

PR #27, opened 2024-03-06
feat: add "in cellxgene" column determined by doi matches (#16)

hunterckx:


files/build-datasets.ts
```diff
@@ -136,34 +138,64 @@ export async function buildAtlasDatasets(
...
+        "disease"
+      ),
+      estimatedCellCount: calculateEstimatedCellCount(projectsResponse),
+      inCellxGene: getBooleanLabel(Boolean(cxgCollection)),
+      libraryConstructionMethod: processAggregatedOrArrayValue(
+        projectsResponse.protocols,
+        "libraryConstructionApproach"
+      ),
+      projectId:
+        processEntityValue(projectsResponse.projects, "projectId") || null,
```
frano-m, line 166: The function `processEntityValue`, will return "Unspecified", when the `label` is undefined. See [here](https://github.com/clevercanary/hca-atlas-tracker/blob/1b579a3b65b9c8b9f780d7ea05a263b82ab10771/files/apis/azul/utils.ts#L157). If you want `null` as the fallback value for `projectId`, pass in `LABEL.EMPTY`. The returned value will be an empty string which be "false" for the logical OR condition.

frano-m (approved):
Awesome 🥇 @hunterckx.

PR #31, opened 2024-03-07
feat: configure atlas detail page (#21)

### LINKED ISSUE #21: Configure atlas detail page
Similar to data portal atlas detail page but include the source datasets and component atlases both as tabs instead of having the component atlases on the overview page.

Use the columns already in tracker for the tab column definitions.

frano-m:
Closes #21.


NoopDog (approved):
Awesome work @frano-m Thanks! 🚀

PR #47, opened 2024-03-28
test: use separate database for tests (#45)

hunterckx:
- Slight change to existing migration runner (ending the pool explicitly to avoid a delay in the script completing)
- Separate test DB migration runner
- Mock for pg-connect-config so that tests will use the test database


NoopDog (approved):
LGTM! Thx

Issue #43, opened 2024-03-19
Create Atlas flow - minimal


NoopDog:
#### Need
We will need a table to load and save atlases in the database.  We want to be able to load just the table view of the atlas without loading up all the atlas fields.

#### Approach

Create flow

- add post method, logged in only.
- add create atlases service method, protect for content admins.
- add atlases table migration. 
- return the created atlas in the create response.
- response code is 201 entity created

Table: `atlases`

1. id UUID primary key
2. status string, required 
3. overview jsonb
4. source_datasets jsonb
8. created_at timestamp auto set at create
9. updated_at required  auto set on update, set to updated_at on create

Minimal create form input
(add to overview jsonb column)
1. Network (e.g. eye) required in the controller
2. short_name (e.g. retina) required in the controller
3. Version (e.g. 1.0) required in the controller

And tests...

### Create Atlas Flow

1. User selects "Add Atlas" from the /atlases page.
2. System shows the /atlases/create form with fields
   1. Short Name
   2. Version
   3. Network
3. The user submits the form.
4. The front end validates the fields and shows any validation errors.
5. If no validation errors the system submits the form to the back end.
6. The back end validates the input, saves it in the database, and returns an OK response and the created Atlas.
7. When the create form gets a good response, it forwards to /atlases/edit/[id], where the ID is obtained from the create response.
8. The create form can be server-side rendered (see getServerSideProps )for getting the Atlas data on the server)
9. Fin (we can stop here and do additional work on other tickets)

NoopDog:
Complete

PR #49, opened 2024-03-30
feat: create minimal atlas backend and form (#43)

hunterckx:
~~Question: should the atlas overview correspond more closely to the existing atlas entity? (e.g. `network` vs `bioNetwork`, and I think `short_name` vs `atlasKey`)~~


pages/atlases/create.tsx
```diff
@@ -0,0 +1,15 @@
+import { GetStaticProps } from "next";
+
+export const getStaticProps: GetStaticProps = async () => {
```
NoopDog, line 3: Do we need to get static props? I don't think this page needs to be pre-rendered. I think we can remove getStatic Props and getStaticPaths for the forms.

pages/api/atlases/create.ts
```diff
@@ -0,0 +1,33 @@
...
+export default handler(
+  method("POST"),
+  role("CONTENT_ADMIN"),
+  async (req, res) => {
+    let newInfo: NewAtlasData;
+    try {
+      newInfo = await newAtlasSchema.validate(req.body);
+    } catch (e) {
+      if (e instanceof ValidationError) {
+        res.status(400).end();
```
NoopDog, line 21: I think we can send the validation errors back in the response in such a way they can be displayed on the form.
We will client-side validate. But it's helpful to have the full error message returned on the server side validation.

migrations/1711747625968_atlas-table.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+    `
+      BEGIN
+        NEW.updated_at = CURRENT_TIMESTAMP;
+        RETURN NEW;
+      END;
+    `
+  );
+
+  pgm.createTrigger({ name: "atlases", schema: "hat" }, "update_updated_at", {
```
NoopDog, line 47: NIce, thx

NoopDog (changes requested):
Thx @hunterckx a few comments to check out. Ping me on slack with any q's or further clarification! Thx

app/apis/catalog/hca-atlas-tracker/common/schema.ts
```diff
@@ -1,16 +1,32 @@
...
+const NETWORK_REGEXP = new RegExp(
+  `^(?:${NETWORK_KEYS.map(escapeRegExp).join("|")})$`
+);
+
+/**
+ * Schema for data used to create a new atlas.
+ */
+export const newAtlasSchema = object({
+  network: string().required().matches(NETWORK_REGEXP),
+  short_name: string().required(),
```
NoopDog, line 14: lets leave these as is, maybe we will change the ones on the atlas pages.

NoopDog (commented):
Thx @hunterckx a few comments to check out. Ping me on slack with any q's or further clarification! Thx

NoopDog (approved):
⭐ ⭐ Awesome 🚀  thanks @hunterckx  ⭐ ⭐

Issue #62, opened 2024-04-05
Add "Wave" to Atlas create and edit flow


NoopDog:
#### Need 

Atlas "Wave" is a rough indicator of priority or order of atlas creation.  Wave can be 1, 2 or 3, with wave one atlases being created first.

<img width="788" alt="image" src="https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/318f23cf-b506-4364-9ba4-73939e7cf23a">

We need to add the wave to the atlas to create flow, end to end, from form to database.

#### Definition of done

- [ ] 1. Wave is styled like the mocks using the newly styled create form as a guide.
- [ ] 2. Wave appears on the Create Atlas form in a dropdown with values 1,2 or 3.
- [ ] 3. Wave is validated client and server side.
- [ ] 4. Wave is saved in the overview JSONB on the atlases table
- [ ] 5. Wave shows on the atlases table
- [ ] 6. Wave is a filter on the altases table
- [ ] 7. Tests are updated to include wave

NoopDog:
complete

PR #88, opened 2024-04-12
feat: change wave to be a string and be initially unspecified on form (#62)

hunterckx:
Also changes the schema to use `oneOf` for network and and wave, meaning the schema is typed more accurately, hence the changes involving type assertions

### DIFF
--- a/app/views/EditAtlasView/hooks/useEditAtlasForm.tsx
+++ b/app/views/EditAtlasView/hooks/useEditAtlasForm.tsx
@@ -30,13 +30,17 @@ export const useEditAtlasForm = (
  * @param atlas - Atlas.
  * @returns schema default values.
  */
-function mapSchemaValues(atlas?: HCAAtlasTrackerAtlas): AtlasEditData {
-  return {
-    [FIELD_NAME_ATLAS_NAME]: atlas?.shortName || "",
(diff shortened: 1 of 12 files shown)


app/views/EditAtlasView/hooks/useEditAtlasForm.tsx
```diff
@@ -30,13 +30,17 @@ export const useEditAtlasForm = (
...
+  atlas?: HCAAtlasTrackerAtlas
+): AtlasEditData | undefined {
+  return (
+    atlas && {
+      [FIELD_NAME_ATLAS_NAME]: atlas.shortName,
+      [FIELD_NAME_BIO_NETWORK]: atlas.bioNetwork,
+      [FIELD_NAME_VERSION]: atlas.version,
+      [FIELD_NAME_WAVE]: atlas.wave,
+    }
+  );
```
hunterckx, line 43: This change was also made due to the schema being typed more strictly, and should, in practice, work the same as 
it did before, since the result just gets passed to `useForm`, which will call `schema.cast` to fill in default values anyway

NoopDog: cc: @frano-m

NoopDog (approved):

Issue #63, opened 2024-04-05
Add "integration lead to create and edit atlas backend


NoopDog:
#### Need
Atlases have an "Integration lead" responsible for coordinating source dataset selection, and atlas creation. The integration lead's name and email address must be added to the atlas. 

<img width="1279" alt="image" src="https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/5ae8f4b9-bba2-479c-9cdb-fdf2e1fb4143">

#### Approach 
The front end for this will be done on a separate ticket. This ticket is concerned with adding the integration lead as optional parameters to the atlas edit API.

`integrationLead: null` or...

```
integrationLead: {
    "name": "Joe Smith",
    "email": "joe@smith.com"
}
````

#### Definition of Done
1. Edit Atlas API takes an optional `integrationLead` object with required fields `name` and `email` This object's fields are validated if the object is present and saved in the overview JSONB table.
2. integration lead is returned in the `/api/atlases[id]` response.

NoopDog:
@frano-m this is merged !

NoopDog:
Complete

PR #95, opened 2024-04-12
feat: add integration lead to create and edit apis (#63)

hunterckx:


app/views/EditAtlasView/hooks/useEditAtlasForm.tsx
```diff
@@ -39,6 +39,7 @@ function mapSchemaValues(
       [FIELD_NAME_BIO_NETWORK]: atlas.bioNetwork,
       [FIELD_NAME_VERSION]: atlas.version,
       [FIELD_NAME_WAVE]: atlas.wave,
+      integrationLead: atlas.integrationLead,
```
hunterckx, line 42: Using the key directly here rather than via a `FIELD_NAME_` constant; my thinking was that it would be nice to avoid unnecessary changes to `generalInfo.tsx`, where the constants are exported from (and which Fran has already edited in her integration lead branch), though maybe this just creates additional work to make things consistent....

(In any case, I question whether the constants are necessary in general)

NoopDog (approved):

Issue #81, opened 2024-04-10
Query the HCA Data Repository to determine if a source dataset has been ingested


NoopDog:
### Need
A key feature of the tracker is to check on the ingestion status of datasets in the various systems that make up the wider HCA Data Ecosystem (HCA Data Repository, CELLxGENE, CAP, DUOS) etc.  When the system supports it, source datasets (and Atlases) will be synchronized by the publication DOI. When a dataset in the external system with a matching DOI is found.

![Image](https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/202aead8-3ec7-4453-9c71-a7708967b05f)

![Image](https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/4fc6751c-5215-41f0-ace5-176b0fba732c)

#### User Flow

1. The user enters a source dataset DOI into the system using the "Create Source Dataset" flow. 
2. The system processes the DOI, checks CrossRef for the DOI and if a matching publication is found, portions of the publication metadata are saved in a new source dataset record.
3. The system then checks if a matching DOI exists in the HCA Data Repository and if so adds the HCA Data Repository project ID to the source dataset record. 
4. The source dataset edit form will show the source dataset found if any.
5. The entry for the source dataset on the Atlas's source dataset list will show the status of the HCA Repository ingest.

We want to be able to check for matching DOIs in real-time (sub-second) so that this check can be included in the form processing flow. 

The HCA Data Repository is updated roughly monthly.  

The HCA Data Repository does not have an API for finding projects by DOI.

### Approach

At a high level, the approach for this can be:

1. Query the /index/catalogs API to determine the default catalog.
2. Create a map of DOI to the project by reading all of the datasets in the default catalog
3. Determine if the map of DOI to projects is stale by calling the /index/catalogs API and seeing if the catalog has changed.
4. Refresh the DOI to project map if the default catalog has changed
5. Query the map by DOI to determine the project ID, if any, and save it on the source dataset. 

#### Default Catalog
The HCA Data Repository content is stable across "Catalogs." There is an [index/catalogs](https://service.azul.data.humancellatlas.org/index/catalogs) API that returns a list of catalogs and an indicator of the `default_catalog.` The `/index/catalogs` response returns fairly quickly (about 300ms). 

Here the default catalog is `dcp35`
[...]

PR #96, opened 2024-04-14
feat: query hca to associate dois with projects (#81)

hunterckx:


app/apis/azul/hca-dcp/common/entities.ts
```diff
@@ -0,0 +1,88 @@
...
+  count: number;
+  format: string;
+  matrixCellCount: number;
+  totalSize: number;
+}
+
+/**
+ * Model of project value in the response from index/projects API endpoint.
+ */
+export interface ProjectResponse {
```
NoopDog, line 43: Can this be imported from the package or is in data-browser?

hunterckx:
Additional notes:
- We want to initialize the DOI-to-project mapping when the server starts, which is why I've enabled Next's "instrumentation" feature, since that allows running a function on server start
- Storing the projects info on the global object (as opposed to using a singleton pattern) is what makes the data accessible after the instrumentation runs and between rebuilds done by the dev server

app/viewModelBuilders/catalog/hca-atlas-tracker/common/viewModelBuilders.ts
```diff
@@ -324,3 +325,38 @@ function getSourceDatasetPublicationColumnDef(): ColumnDef<HCAAtlasTrackerSource
     header: HCA_ATLAS_TRACKER_CATEGORY_LABEL.PUBLICATION,
   };
 }
+
+/**
+ * Get task status describing whether a source dataset is known to be in CAP.
+ * @param sourceDataset - Source dataset.
+ * @returns whether the source dataset is in CAP.
+ */
+function getSourceDatasetInCap(
```
frano-m, line 334: I wonder if it is worth making each of these a single function and we pass in the key `capId` or `cellxgeneCollectionId` etc to grab the value, and then return DONE or REQUIRED?

Or, do you anticipate the returned values could become more complex and unable to be generalised?

app/utils/hca-projects.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+  return projectIdsByDoi;
+}
+
+/**
+ * Fetch current default HCA catalog name.
+ * @returns catalog name.
+ */
+async function getLatestCatalog(): Promise<string> {
```
frano-m, line 86: @hunterckx are we able to utilise `fetchCatalog` from [findable](https://github.com/clevercanary/data-explorer/blob/main/packages/data-explorer-ui/src/entity/api/service.ts#L86)?

frano-m (approved):
Excellent work @hunterckx thank you ⭐ 🚀

Issue #113, opened 2024-04-17
Match DOIs by equality,  has-preprint, or is-preprint-of


NoopDog:
#### Need
A dataset may be ingested into a system when it is a preprint; later, the dataset is referenced by its publication URL. For example, see 10.1073/pnas.1908706116.

 When searching for a DOI in an external system, we can consider the preprint the DOI was derived from or the publication derived from the preprint as also matching. 

#### Approach
When looking up DOIs, examine the has-preprint or is-preprint-of fields to extract the related DOIs.

```
"relation": {
      "has-preprint": [
        {
          "id-type": "doi",
          "id": "10.1101/645424",
          "asserted-by": "object"
        }
      ],
```

```
"relation": {
      "is-preprint-of": [
        {
          "id-type": "doi",
          "id": "10.1073/pnas.1908706116",
          "asserted-by": "subject"
        }
      ]
    },
```
If both a publication and preprint DOI are found, search for matches first using the publication DOI, then try using the preprint DOI.

#### Definition of Done
1. isPubicationOf or isPrePrintOf DOIs are saved on the source dataset along with the given DOI in the crossref section.
2. Searches for DOIs in the DOI caches try to match by the publication DOI and then by the preprint DOI.
3. The project match is made if the HCA Repository only has the preprint but the user enters the publication.
4. The reverse of 3 is true!
5. Tests for these cases.

NoopDog:
Complete

Issue #86, opened 2024-04-11
Handle "preprint" subtype in CrossRef API response


NoopDog:
### Need
Crossref has a "subtype" in their publication metadata:

![Image](https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/c5f2f62b-d6d0-4267-8198-b0e69280c2db)

The main subtypes we are interested in are "preprint" and "journal". We will need to see if the DOIs we see have different subtypes, but for now, this is the set of subtypes I can see is getting. 

When the "preprint" subtype is selected, there is no "journal" specified so we can't build our publication string.  We need to detect preprints and update the formation of the publication string to handle the preprint case.

### Approach

Pass through the subtype form crossref through to the UI so we can use it to help format the publication string and the source dataset display.

#### DOD

- [ ] 1. Add the publication subtype to the source database database entry.
- [ ] 2. Add the publication subtype to the source dataset API response
- [ ] 3. If the publication subtype is something other than "journal" or "preprint," reject the form and submit with an error "Unsupported publication subtype: ${subtype} on the DOI field. 
- [ ] 4. If the publication subtype is a preprint, and there's no `journal` value from `container-title`, `short-container-title`, or `institution`, set `journal` to `Preprint`. show the publication string like `Choudhury (2020) Preprint` on the list and create and update forms otherwise, use the preprint publication as normal like `Choudhury (2020) Bioarxiv`.
- [ ] 5. Tests for unsupported, journal and preprint subtypes API responses and publication string builder.

frano-m:
LGTM @hunterckx thank you! 🚀

NoopDog:
Complete

PR #124, opened 2024-04-18
feat: validate crossref responses and give preprints default journal value "Preprint" (#86)

hunterckx:
Notes:
- The information from the ticket about Crossref's `subtype` property seems to be inaccurate -- preprints do have the `preprint` subtype, but from what I've seen journal articles have a `type` of `journal-article` and no subtype
- Since we updated the ticket to specify "Preprint" as a value for the `journal` property rather than something inserted when the publication string is built, I haven't included the preprint-vs-journal information on the source datasets. (If this is something we want after all, maybe it it can be included as part of #113?)
- A notable part of this not addressed much in the ticket is that error responses from APIs may now have a JSON body containing an `errors` object that maps field names to arrays of error messages, which are displayed as errors on the forms


app/apis/catalog/hca-atlas-tracker/common/entities.ts
```diff
@@ -117,7 +117,7 @@ export type Wave = (typeof WAVES)[number];
 
 export interface PublicationInfo {
   authors: Author[];
-  journal: string | null;
+  journal: string;
```
hunterckx, line 120: `journal` is now required in PublicationInfo

app/utils/crossref/crossref-api.ts
hunterckx, line 1: This is separate from `crossref.ts` so that `fetchCrossrefWork` can be mocked for tests while keeping the actual definition of `getCrossrefPublicationInfo`

app/hooks/useForm/useForm.ts
```diff
@@ -39,7 +48,20 @@ export const useForm = <T extends FieldValues, R = undefined>(
...
   const [data, setData] = useState<R | undefined>();
   const [submitDisabled, setSubmitDisabled] = useState<boolean>(false);
-  const { reset } = formMethod;
+  const { reset, setError } = formMethod;
+
+  const onError = useCallback(
+    (errors: FormResponseErrors) => {
+      if ("message" in errors) throw new Error(errors.message); // TODO display these errors?
+      for (const [field, messages] of Object.entries(errors.errors)) {
+        setError(field as Path<YupValidatedFormValues<T>>, {
```
hunterckx, line 57: Because of how yup and react-hook-form work, I can't see a way of making the field names work with Typescript without making some kind of assumption about their validity, and I think this is a reasonable place to do that

frano-m (approved):
LGTM @hunterckx thank you! 🚀

Issue #99, opened 2024-04-14
Query CELLxGENE API to determine the collection matching a source dataset if any


NoopDog:
#### Need
Like https://github.com/clevercanary/hca-atlas-tracker/issues/81, we need to find the entry in CELLxGENE for a given source dataset and can do that by DOI.

#### Approach
This can be slightly different as CELLxGENE updates possibly multiple times a day.  

1. Call the CELLxGENE /collections API on system startup to create a map of the source dataset to a collection.
2. When creating or updating a source dataset, inspect the CELLxGENE map to find a collection ID to add to the source dataset record.
3. if an established DOI pair is missing during the update, remove the mapping from the source dataset. 
4. Consider the CELLxGENE map invalid after 4 hours.
5. Use the current map but kick off a map refresh (SWR style) if the map is older than 4 hours when a request comes in to create or update a source dataset

#### DOD

After creating or updating a source dataset

- [ ] The update form shows the CELLxGENE collection id if any
- [ ] The source dataset list indicates if the match was found
- [ ] Tests

NoopDog:
Complete

PR #128, opened 2024-04-22
feat: query cellxgene to match collections with source datasets (#99)

hunterckx:
This does not implement item 3 of the approach described in the ticket (removing mappings from source datasets)


app/services/hca-projects.ts
hunterckx, line 1: `hca-projects.ts` has been moved from the `utils` folder to a new `services` folder, where the new CELLxGENE service and shared refresh service code also are

NoopDog (approved):

Issue #131, opened 2024-04-22
Create unpublished source dataset backend


NoopDog:
#### Need
See https://github.com/orgs/clevercanary/projects/2/views/11?pane=issue&itemId=60500637
For the front end...

#### Approach
Update the backend /api/atlases/id/source-dataset/create implementation to work with a first author, email and working title instead of the DOI.

#### DOD
1. /api/atlases/id/source-dataset/create tasks FirstAuthorSurname and Year instead of a DOI.
2. If DOI is supplied, unpublished fields are not allowed.
3. If DOI is missing all unpublished fields are required.
4. If there is no DOI, don't call crossref; instead, save unpublished fields on the source dataset.
5. Update the PUBLICATION_STATUS enum to have a "UNPUBLISHED" entry and use this value when there is no DOI?
6. Add an "unpublishedInfo" field and type to hold the unpublished info (Author, email, and working title) This is only populated in the "unpublished" case.
7. On the source dataset list, show the working title as the title.
8. On the source dataset list for the publication column, show the author name and email address followed by unpublished like  `Dan Smith, dan@smith.com - Unpublished`

NoopDog:
Complete!

PR #142, opened 2024-04-25
feat: add backend support for unpublished source datasets (#131)

hunterckx:
Notes:
- The schema and how it fits in with the form might need some attention
  - Currently (with the new fields and no update to the form) the save button for some reason doesn't do anything
- API response datasets are flat objects with several properties used for both published and unpublished source datasets
- `firstAuthorPrimaryName` is renamed to `referenceAuthor` (does this seem appropriate?)
- Field names for new unpublished source dataset are ~~`author`~~ `referenceAuthor`, `contactEmail`, and `title`
  - Would `authorEmail` be more accurate? (I'm inclined not to use just `email` because I think it should be more specific on the API response and ideally that would be consistent with the request)
- ~~The only API update is the `create` API (do we need an API for updating source datasets for this ticket?)~~
  - There is now an edit API

EDIT: the API schema accepts either an object of the form `{ doi: string }`, or of the form `{ contactEmail: string; referenceAuthor: string; title: string }` (~~though I guess we may change `author` to `referenceAuthor`?~~ done!). If DOI is a string, the other three properties must be `undefined`/omitted, and if DOI is `undefined`/omitted, the other three properties must be strings.

### DIFF
--- a/app/services/database.ts
+++ b/app/services/database.ts
@@ -1,0 +1,21 @@
+import pg from "pg";
+import { getPoolConfig } from "../utils/pg-app-connect-config";
+
+const { Pool } = pg;
+
+const pool = new Pool(getPoolConfig());
+
+export function query<T extends pg.QueryResultRow>(
+  queryTextOrConfig: string | pg.QueryConfig<unknown[]>,
+  values?: unknown[] | undefined
+): Promise<pg.QueryResult<T>> {
+  return pool.query<T>(queryTextOrConfig, values);
+}
+
+export function getPoolClient(): Promise<pg.PoolClient> {
+  return pool.connect();
+}
+
+export function endPgPool(): void {
+  pool.end();
+}
--- a/app/services/source-datasets.ts
+++ b/app/services/source-datasets.ts
@@ -1,6 +1,8 @@
 import { ValidationError } from "yup";
 import {
+  ATLAS_STATUS,
   DOI_STATUS,
+  HCAAtlasTrackerDBAtlas,
   HCAAtlasTrackerDBSourceDataset,
   HCAAtlasTrackerDBSourceDatasetMinimumColumns,
 } from "../apis/catalog/hca-atlas-tracker/common/entities";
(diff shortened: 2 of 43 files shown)


NoopDog:
@hunterckx 

1. can you document the schema for the POST?
2. Can you implement an update API (PUT) for this?
3. Looks like this PR needs a rebase

app/services/database.ts
hunterckx, line 1: Basic database interactions are moved to their own file, which causes some changes in otherwise unrelated files

app/utils/api-handler.ts
```diff
@@ -17,22 +14,32 @@ export type MiddlewareFunction = (
...
-      await f(req, res, () => (done = false));
-      if (done) return;
+    try {
+      for (const f of funcs) {
+        let done = true;
+        await f(req, res, () => (done = false));
+        if (done) return;
+      }
+    } catch (e) {
+      respondError(res, e);
```
hunterckx, line 42: Handlers created by `handler` and `handleByMethod` now automatically handle errors, allowing the functions in `app/services/source-datasets.ts` to throw appropriate errors and have them translate to responses from the API

app/views/EditSourceDatasetView/common/schema.ts
hunterckx, line 1: I've copied the old versions of the source dataset schemas to be used by the forms, so that the forms still work with the changes to the API schemas

app/services/source-datasets.ts
hunterckx, line 1: `app/services/source-datasets.ts` contains code shared between source dataset creation and editing

pages/api/atlases/[atlasId]/source-datasets/[sdId].ts
```diff
@@ -0,0 +1,74 @@
...
+import {
+  AccessError,
+  getUserRoleFromAuthorization,
+  handleByMethod,
+  handler,
+  NotFoundError,
+  role,
+} from "../../../../../app/utils/api-handler";
+
+async function getHandler(
```
Comment (hunterckx) [inline pages/api/atlases/[atlasId]/source-datasets/[sdId].ts:23]: The new API route, `/api/atlases/[atlasId]/source-datasets/[sdId]`, has a GET handler as well

NoopDog (approved):
Epic! Thanks @hunterckx LGTM!

Issue #140, opened 2024-04-24
Add logging, retries to attempts to refresh the HCA and CELLxGENE id caches


NoopDog:
#### Need
We have an issue where, on the dev deployment, and I presume the prod deployment, we can't get the the cellxgene collections map loaded. We get a timeout calling the cellxgene API (at least, I think that is what is timing out)

Alex says there is no allow list for outbound connections. I don't know why HCA would work but CELLxGENE does not. I'm trying to remember if there is a difference in how/when they are triggered. I have a theory that the server comes up and starts booting node before the network is fully available. 

#### Approach
Let's add logging of the attempt, any errors or success, and retry after a delay of a minute  for both the HCA and CELLxGENE caches. Lets do this on the initial attempt on startup and any subsequent attempts.

NoopDog:
Complete

PR #143, opened 2024-04-25
feat: add logging and retries to hca/cellxgene refreshes (#140)

hunterckx:
Retries are done via `ky`, which is made by the same people as `got` but is supposedly preferable because it uses `fetch`


NoopDog (approved):
Thanks @hunterckx 🚀

PR #169, opened 2024-05-03
feat: updates to rules for discard and save header (#133)

### LINKED ISSUE #133: Updates to rules for discard and save header
#### Need

I worked out with Kaspars some updates to the discard and save header rules that smooth things out a bit. They are!:

1. On create, the unsaved changes,  discard, and save header shows. The difference from the current is we show the "Unsaved changes" even if the form has not been touched yet.
2. On viewing the detail (edit mode with no dirty fields) the discard and save header does NOT show. - This is a change from the current implementation.
3. On the dirty of the edit form, the header appears.
4. On save, the header disappears as there are no longer any dirty fields.
5. This applies to both atlases and datasets.
6. If the user has a dirty atlas form, and tries to go to any of the other tabs, they get a popup telling them they have unsaved changes

https://www.figma.com/file/i4uZsYkKAML9ivi2TIo6ug/HCA-Data-Explorer?type=design&node-id=7510%3A8938&mode=design&t=XSnjLHvYi3jXOVA0-1

Saving saves changes and shows the new tab.
Discard, discards the changes, and shows the new tab.
X just closes the box and does not save or discard.

image

frano-m:
Closes #133.


NoopDog (approved):
Amaze! 👍

Issue #177, opened 2024-05-03
Console log validations on create and edit for source datasets


NoopDog:

NoopDog:
Complete!

PR #187, opened 2024-05-07
feat: log validations on create/update source dataset (#177)

hunterckx:


app/services/validations.ts
```diff
@@ -0,0 +1,133 @@
...
+  return {
+    atlasIds,
+    description: validation.description,
+    entityId: entity.id,
+    entityTitle,
+    entityType,
+    system: validation.system,
+    taskStatus:
+      validationStatus === VALIDATION_STATUS.VALID
+        ? TASK_STATUS.DONE
```
NoopDog, line 88: Yes no way to tell "IN_PROGRESS" at the moment so leave it out.

app/services/validations.ts
```diff
@@ -0,0 +1,133 @@
...
+    sourceDataset.sd_info.unpublishedInfo?.title ??
+    sourceDataset.id
+  );
+}
+
+async function getSourceDatasetAtlasIds(
+  sourceDataset: HCAAtlasTrackerDBSourceDataset
+): Promise<string[]> {
+  const queryResult = await query<Pick<HCAAtlasTrackerDBAtlas, "id">>(
+    "SELECT id FROM hat.atlases WHERE source_datasets @> $1",
```
NoopDog, line 129: I wonder how we can add an index to support this query. We may not really need it at this scale but would be good to have.

app/services/validations.ts
```diff
@@ -0,0 +1,133 @@
...
+  };
+}
+
+export async function updateSourceDatasetValidations(
+  sourceDataset: HCAAtlasTrackerDBSourceDataset
+): Promise<void> {
+  const title = getSourceDatasetTitle(sourceDataset);
+  const atlasIds = await getSourceDatasetAtlasIds(sourceDataset);
+  for (const validation of SOURCE_DATASET_VALIDATIONS) {
+    if (validation.condition && !validation.condition(sourceDataset)) continue;
```
NoopDog, line 102: simple! Thx

pages/api/atlases/[atlasId]/source-datasets/[sdId].ts
```diff
@@ -59,6 +60,7 @@ const putHandler = handler(
       sdId,
       await sourceDatasetEditSchema.validate(req.body)
     );
+    await updateSourceDatasetValidations(newDataset);
```
Comment (NoopDog) [inline pages/api/atlases/[atlasId]/source-datasets/[sdId].ts:63]: Place in a transaction with updateSourceDataset. Also, I think this call should be moved to the dataset service. Let the service sequence the database calls, not the "handler."

NoopDog (approved):

NoopDog:
Awesome! Thanks @hunterckx

PR #188, opened 2024-05-07
feat: only show the create, edit buttons and forms if the user is a content editor (#138)

### LINKED ISSUE #138: Only show the create, edit buttons and forms if the user is a Content Editor
#### Need
Currently, the create and edit buttons and forms for atlas and source datasets show if the user is logged in but not registered. Instead, we need to show these forms only if the user has the role CONTENT_ADMIN.

#### Approach
1. Call the new[ /me](https://github.com/clevercanary/hca-atlas-tracker/issues/144) endpoint to determine the user's role after login.
2. Only show the create/update atlas and source dataset buttons if the user is a CONTENT_ADMIN.

frano-m:
Closes #138.


app/components/Detail/components/TrackerForm/components/Section/components/Atlas/components/GeneralInfo/generalInfo.tsx
```diff
@@ -27,14 +28,24 @@ import {
...
+  formMethod,
+}: GeneralInfoProps): JSX.Element => {
+  const {
+    access: { canEdit },
+  } = formManager;
+  const {
+    control,
+    formState: { errors },
+  } = formMethod;
+  const readOnly = !canEdit;
```
NoopDog, line 48: Perhaps move into form manager so the form manager can explicitly tell you if the form is read-only, and you can avoid repeating  `const readOnly = !canEdit;` in each form.

app/hooks/useFetchUser.ts
```diff
@@ -0,0 +1,41 @@
...
+      accessToken: string
+    ): Promise<HCAAtlasTrackerActiveUser | undefined> => {
+      const res = await fetch(
+        API.USER,
+        getFetchOptions(METHOD.GET, accessToken)
+      );
+      if (isFetchStatusOk(res.status)) {
+        return await res.json();
+      }
+      // TODO handle error.
```
NoopDog, line 24: Why is this TODO handle error? Is the error not already handled below?

app/components/common/Form/components/FormManager/components/Popup/popup.tsx
```diff
@@ -15,15 +15,22 @@ interface PopupProps extends FormManager {
...
+  const { canEdit } = access;
   const { onCancel, onDiscard, onSave } = formAction || {};
   const { isLeaving } = formStatus;
   return (
-    <Dialog fullWidth maxWidth="xs" onClose={onCancel} open={isLeaving}>
+    <Dialog
+      fullWidth
+      maxWidth="xs"
+      onClose={onCancel}
+      open={canEdit && isLeaving}
```
NoopDog, line 32: Let's see. Why does the dialog need canEdit? The form can never be dirty unless the user is an editor. Does the content of the dialog change if the user is an editor? I'm not sure how a non-editor would ever see this dialog but I am not sure about the full implications of the flow.

NoopDog (approved):

app/components/Detail/components/TrackerForm/components/Section/components/Atlas/components/GeneralInfo/generalInfo.tsx
frano-m: True, I contemplated this - I wasn't sure if the `readyOnly` would always be determined by the `canEdit` in future iterations...

app/hooks/useFetchUser.ts
frano-m: A copy and paste job - sorry :)

PR #200, opened 2024-05-13
feat: sort by publication ascending on source dataset list (#154)

### LINKED ISSUE #154: Sort by publication ascending on source dataset list
#### Need
The source datasets are almost always given by something similar to the journal string in external notes.
We need to make it easier to tell if, for example, "Young et al." is included in the source datasets list.

#### Approach
Update the default sort of the source datasets column in the config.

#### DOD

1. Order by Publication ascending 

That's it!

frano-m:
Closes #154.

Sorts source datasets by citation, and then by title (should the citation be the same).

image


NoopDog (approved):
Sorted! 😀

Issue #183, opened 2024-05-05
Re-validate all source datasets when the refresh service updates


NoopDog:

NoopDog:
Complete!

PR #208, opened 2024-05-19
feat: revalidate all source datasets when refresh services finish refreshing (#183)

hunterckx:
Notes:
- Revalidation happens when either of the refresh services finishes refreshing and there are no other refreshes in progress -- meaning that when both HCA and CELLxGENE get refreshed at the same time, whichever finishes last will trigger revalidation
- Because the server does an initial refresh, revalidation will always happen when the server is started


NoopDog (approved):
Ok then! Looks great. Thx

Issue #207, opened 2024-05-18
Add API endpoint to set task target completion date


NoopDog:
### Need
See the [front-end ticket](https://github.com/clevercanary/hca-atlas-tracker/issues/205).

### DOD

1. PATCH endpoint to support bulk update of task completion dates.
2. The targetCompletion date should be in ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ Example: 2024-05-17T10:00:00Z . The front end will map dates to quarters or months etc. for display. 
3. The command fails with 400 if the date is not valid.
4. The command fails if one of the IDs is invalid with 404.
5. Tests!

### Technical Design

Below is an initial approach with some code sketches to accomplish the above, extracted from this [ChatGPT session](https://chatgpt.com/c/58b3246c-6c2d-48b8-8dca-22ef03cfbaea).

**API Endpoint:** `/api/tasks/completionDates`

**HTTP Method:** `PATCH`

## Request

- **Endpoint:** `/api/tasks/completionDates`
- **Method:** `PATCH`
- **Request Body:**
  ```json
  {
    "taskIds": ["id1", "id2", "id3"],
    "targetCompletion": "2024-05-17T10:00:00Z"
  }
  ```

## Response

- **Status Codes:**
  - **200 OK**: The update is successful, and the response includes the updated entities.
  - **400 Bad Request**: The request is malformed or contains invalid data. e.g. the date is not a valid date.
  - **404 Not Found**: One or more of the specified task IDs do not exist.

- **Response Body (200 OK):**
  ```json
  [
    {
      "id": "id1",
      "targetCompletion": "2024-05-17T10:00:00Z",
      "updatedAt": "2024-05-17T12:00:00Z",
      "otherField1": "value1",
      "otherField2": "value2"
    },
    {
      "id": "id2",
      "targetCompletion": "2024-05-17T10:00:00Z",
      "updatedAt": "2024-05-17T12:00:00Z",
      "otherField1": "value1",
      "otherField2": "value2"
    },
    {
      "id": "id3",
      "targetCompletion": "2024-05-17T10:00:00Z",
      "updatedAt": "2024-05-17T12:00:00Z",
      "otherField1": "value1",
      "otherField2": "value2"
    }
  ]
  ```

  We want to return the full updated task objects in the response.

## SQL Update Statement

The SQL update statement must ensure that if one of the provided IDs is not valid, the request should be rejected with a `404 Not Found` error.

```sql
[...]

PR #209, opened 2024-05-20
feat: add api to update target completion dates for tasks (#207)

hunterckx:
The endpoint is `/api/tasks/completion-dates` rather than the `/api/tasks/completionDates` specified in the ticket, in order to follow the established naming patterns


NoopDog (approved):
🚀 nice! Thx

Issue #180, opened 2024-05-05
Validate HCA Data Repository Projects for source datasets contain primary data (FASTQ files)


NoopDog:
#### Need
The HCA Data Repository is meant to hold the primary data (FASTQ) files associated with the source datasets. We will validate that this is true and report on it as one of our validations.

#### Approach
Update the refresh service to indicate if the HCA Data Repository project has primary data.  This can be done by examining the Azul projects response and looking at the `fileTypeSummaries` of a project response hit and looking for `format` to be `fastq` or `fastq.gz`.

For example

![Image](https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/19543c03-d2d5-4732-9b58-c7cecf6d544e)

NoopDog:
Complete, Thanks!

PR #215, opened 2024-05-21
feat: add validation for source dataset projects containing primary data (#180)

hunterckx:


app/services/validations.ts
```diff
@@ -70,27 +72,55 @@ export const SOURCE_DATASET_VALIDATIONS: ValidationDefinition<HCAAtlasTrackerDBS
...
+      description: "Add primary data.",
+      system: SYSTEM.HCA_DATA_REPOSITORY,
+      validate(sourceDataset): boolean | null {
+        return validateSourceDatasetHcaProjectInfo(
+          sourceDataset,
+          (projectInfo) => Boolean(projectInfo?.hasPrimaryData)
+        );
+      },
+      validationId: VALIDATION_ID.SOURCE_DATASET_HCA_PROJECT_HAS_PRIMARY_DATA,
+      validationType: VALIDATION_TYPE.INGEST,
```
hunterckx, line 94: Do the properties here for the new validation look good? I'm actually thinking now that I may have used the wrong validation type but I'm not totally certain what the difference between ingest and metadata validations is

NoopDog (approved):
🆒 🫘 !

Issue #217, opened 2024-05-21
Check HCA Title validation is correct


NoopDog:
This DOI when processed shows that the title is incorrect in HCA

https://doi.org/10.1016/j.immuni.2023.01.002

However, I am seeing the titles match HCA and the publication:

The landscape of immune dysregulation in Crohn's disease revealed through single-cell transcriptomic profiling in the ileum and colon
The landscape of immune dysregulation in Crohn’s disease revealed through single-cell transcriptomic profiling in the ileum and colon

Can you determine why we see the validation error?

![Image](https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/75a0325f-a097-4773-9660-d1695a1814eb)

NoopDog:
Additionally, this validation is failing on.. a + vs a superscript +. I imagine we would allow that and that there will be a few similar cases we need to adjust for.

https://doi.org/10.1073/pnas.2204269120

NoopDog:
Don't worry about adjusting for the superscript + for a bit. I want to check if we can fix these automatically.

PR #226, opened 2024-05-23
feat: use looser comparison for title validation (#217)

hunterckx:
Before comparison:
- HTML is removed*
- Unicode characters are normalized
- Case is lowered
- Punctuation and symbols are removed
- Sequences of whitespace are replaced with single spaces
- Beginning/ending whitespace is trimmed

\* Using isomorphic-dompurify -- not certain that's the best option since it's for producing sanitized HTML, not converting to plain text, but it's something we already had installed. Other options might include [jsdom](https://www.npmjs.com/package/jsdom) (which actually is already used by isomorphic-dompurify), [html-to-text](https://www.npmjs.com/package/html-to-text), or [cheerio](https://www.npmjs.com/package/cheerio)


NoopDog (approved):
Yup looks great ! Thx

Issue #220, opened 2024-05-22
Add edit source dataset API and allow updating the CAP ID


NoopDog:
### Need

CAP will start ingesting datasets soon and we ant the tracker to track this and hold the ID for the source datasets ingested in to CAP.

![Image](https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/b402ac24-e74d-4395-8207-3c5288c2bd71)

As CAP has no API currently, we will need to allow for manual input of the ID and post this when submitting the source dataset edit form.

#### Approach

1. Determine the API to use for editing a source dataset and the allowed values (currently only CAP ID, but other fields will come) probably like PATCH `/api/source-datasets/id`.
3. Allow the patch to set or clear the CAP ID.
4. Re-validate the source dataset on update.
5. Return the CAP ID on the `/api/source-dataset/id` and `/api/atlas/id/source-datasets` APIs similar to how the other IDs are returned.

This needs to be compared to the reality of what the edit source datasets API already does!

NoopDog:
Complete!

PR #228, opened 2024-05-24
feat: add cap id to source dataset model and edit api (#220)

hunterckx:


app/apis/catalog/hca-atlas-tracker/common/schema.ts
```diff
@@ -38,6 +47,59 @@ export const atlasEditSchema = newAtlasSchema;
 
 export type AtlasEditData = NewAtlasData;
 
+/**
+ * Create schema that combines an unpublished source dataset schema and a published source dataset schema.
+ * @param publishedSchema - Published source dataset schema.
+ * @param unpublishedSchema - Unpublished source dataset schema.
+ * @returns schema that validates both published and unpublished source datasets.
+ */
+function makeSourceDatasetUnionSchema<T extends { [k: string]: unknown }>(
```
NoopDog, line 56: It is possible that unpublished datasets are ingested into CAP, HCA, and CELLxGENE. Would changes here need to be made to support that?

hunterckx: The API and schema would need to be updated if we want to support setting those IDs for unpublished source datasets, yeah

NoopDog (approved):
OK then!  Nice work @hunterckx 🚀🚀🚀🚀🚀

Issue #213, opened 2024-05-20
Allow for multiple "Integration Leads" on an atlas


NoopDog:
#### Need
Several atlases have two integration leads. We need to accommodate multiple.

#### Approach

1. Update the database so that "Integration Lead" is an array of integration lead objects with a DB migration that reads and writes back the updated format 
2. Update the API response so that integrationLeads is also an array.
3. Update the FE to handle the updated shape.
4. Add Integration lead input on the atlas to create and view detail forms.
5. Show integration leads in alpha order on the atlas forms. 
6. For now, we can support only two leads on the UI.

NoopDog:
Complete!

PR #235, opened 2024-05-25
feat: allow multiple integration leads (#213)

hunterckx:


app/hooks/useForm/useForm.ts
```diff
@@ -38,7 +39,8 @@ export interface UseForm<T extends FieldValues, R = undefined>
 export const useForm = <T extends FieldValues, R = undefined>(
   schema: ObjectSchema<T>,
   apiData?: R,
-  mapSchemaValues?: MapSchemaValuesFn<T, R>
+  mapSchemaValues?: MapSchemaValuesFn<T, R>,
+  mapApiValues: MapApiValuesFn<T> = (p): unknown => p
```
hunterckx, line 43: In order to get the integration leads into an array, I've added an optional `mapApiValues` function to `useForm`, which was tricky to type while allowing it to be left unspecified; what I've done here to make it work is to give it a return type of `unknown`, which works because the return value is just passed to the API which doesn't involve type checks. (I'm not sure how we would want this to work in the future, but this is a fairly minimal change that makes our current setup for multiple integration leads work)

NoopDog (approved):
Yup, Awesome. LGTM!

Issue #234, opened 2024-05-25
Update Atlas filter on "Tasks" list to filter by atlas + version


NoopDog:
Currently, on the "Tasks" tab, while the "Atlas" column shows short name + version, the filter "Atlas" only shows the short name. 

<img width="854" alt="image" src="https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/c6bf8628-6800-4708-9c0b-5ae674432922">

Update the Atlas filter so that it allows filtering on atlas+version in the same manner that the Atlas tab does.

<img width="1020" alt="image" src="https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/bebf7de2-4b04-41b3-b6d1-7e7ac28daec7">

Name both table columns and filters "Atlas" for consistency across the "Atlases" and "Tasks" tabs.

#### Approach
Check how the Atlas table does this. There may be a hidden column to filter by. 

#### DOD

- [ ] Tasks table  is filterable by Atlas version
- [ ] Tasks and Atlases have matching "Atlas" column and filter to filter on Atlas + version.

NoopDog:
Complete!

PR #236, opened 2024-05-25
feat: update atlases and tasks tables to have consistent versioned "Atlas" filter (#234)

hunterckx:


app/services/validations.ts
```diff
@@ -184,6 +184,7 @@ export async function getValidationRecords(
         `
           SELECT
             v.*,
+            ARRAY_AGG(DISTINCT concat(a.overview->>'shortName', ' v', a.overview->>'version')) AS atlas_names,
```
hunterckx, line 187: In order to match up versions with short names, the atlas names for a validations are constructed as part of the SQL query

NoopDog (approved):
LGTM! Nce work!

Issue #230, opened 2024-05-24
Add target completion to the atlas model , APIs and Atlas entity list


NoopDog:
#### Need
See [epic](https://github.com/clevercanary/hca-atlas-tracker/issues/229).

#### Approach

1. Add targetCompletion to the atlas model and database similarly to how it was added to the Tasks table and model.
2. Allow targetCompletion to be set (optional) on the edit atlas (PATCH) API
3. Return the targetCompletion on the Atlas API response and in the Atlas list and detail APIs
4. Add "Target Completion"  as the last item in the atlas entity list UI, similar to tasks with "Unplanned" if no date

NoopDog:
Complete

PR #242, opened 2024-05-28
feat: add target completion to atlas model/apis/table (#230)

hunterckx:
Notes:
- Target completion may be set to null or omitted entirely in the API payload, both of which set it to null in the database
- The property name used for responses from the tasks API was `targetCompletionDate`, which was inconsistent with the payload of the task target completion API (where it's `targetCompletion`), with the database (`target_completion`), and this ticket (also `targetCompletion`), so I've updated the tasks response to use `targetCompletion` as well


NoopDog (approved):
LGTM! Thx @hunterckx

Issue #210, opened 2024-05-20
Add Source datasets count column to Atlases table


NoopDog:
#### Need
Show the number of source datasets on the main atlas list. Make the count clickable and link to the source datasets tab for that atlas. Make sure the click target is big enough to be easily clicked on.  Add the source datasets column after "Wave".

<img width="1322" alt="image" src="https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/c91ba158-e9e3-4456-b321-d22153e259bc">

NoopDog:
Complete, thx!

PR #243, opened 2024-05-29
feat: add source dataset count to atlas table (#210)

hunterckx:
Not sure what qualifies as a big enough click target, but I've left the link at its default size and it seems easy enough for me


NoopDog (approved):
Too easy :)

PR #250, opened 2024-05-30
fix: authorizationprovider should mount once (#249)

### LINKED ISSUE #249: AuthorizationProvider should mount once and not remount between the atlases and atlas pages

frano-m:
Closes #249.


NoopDog (approved):
Nice work @frano-m !

PR #260, opened 2024-05-31
feat: link atlas task counts to filtered tasks (#248)

### LINKED ISSUE #248: Link completion indicator on atlases (3/5) to the atlases task list on the task page
#### Need
We need an easy way to see the details of the task completion status for an atlas.

#### Approach
Make the (4/5) indicator link to the atlases tab and select the given atlas e.g. Selecting gut would generate a link like.
https://tracker.data.humancellatlas.org/tasks?filter=%5B%7B%22categoryKey%22%3A%22atlasNames%22%2C%22value%22%3A%5B%22GDN+v1.0%22%5D%7D%5D

Which selects the Atlas filter to be GND v1.0.

It is currently not possible to put the sort order in a link so we can skip that for now.

hunterckx:
Closes #248.

### DIFF
--- a/app/components/Table/components/TableCell/components/TaskCountsCell/taskCountsCell.styles.ts
+++ b/app/components/Table/components/TableCell/components/TaskCountsCell/taskCountsCell.styles.ts
@@ -1,0 +1,13 @@
+import styled from "@emotion/styled";
+
+export const Cell = styled.div`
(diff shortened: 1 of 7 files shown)


NoopDog (approved):
Well done team! ✋

Issue #252, opened 2024-05-30
Tasks and atlas table updates


NoopDog:
### Tasks Table
**Changes to the tasks table and filter**

1. Add DOI as an optional column. Use "Unpublished" if there is no DOI.
2. Add DOI as a filter
3. Add Entity Title as a filter.  call it Title.
4. Make the atlas table element be a hyperlink to the atlas source datasets page.

**Add named groups to the left-hand filters**

Leave "Saved Filters" where it is at the top.

Add filter groups to match

**Task**

3. Status
4. System
5. Target Completion
6. Task (Rename to Task Type)

**Atlas**
1. Atlas
2. Biological Network
3. Wave

**Dataset**
1. DOI
2. Publication
3. Title

### Atlas Table

Update the left-hand filter to have a named filter group like:

**Atlas**
1. Atlas
3. Biological Network
4. Integration Lead
5. Status
6. Target Completion
8. Wave

PR #261, opened 2024-05-31
feat: update filter groups and task columns (#252)

hunterckx:
Not certain about the task entity title link, since tasks can have multiple atlases; currently I have it using the first one in the list


NoopDog (approved):
Ok then! Thx

PR #265, opened 2024-06-03
feat: add cap url to source dataset edit page (#255)

### LINKED ISSUE #255: Add CAP URL to source source dataset edit page
#### Need 

The HCA Atlas creation workflow is to create matrices in CELLxGENE, then ingest these to CAP before a "Annotation Jamboree" is run. To aid in tracking the progress in adding datasets to CAP the tracker will need to know about the CAP dataset link. As there is no current API for CAP we will need to allow the CAP team to input the URLs to the ingested datasets in CAP manually.

#### Approach

Add an input to the source dataset edit form in the identifiers section to take the CAP ID.

[Figma](https://www.figma.com/design/g60pHZLz78wfl0PvbksMb5/HCA-Tracker?node-id=54-3171&t=SIkupXFcLTvOkmjV-1)

image

See the [back-end ticket](https://github.com/clevercanary/hca-atlas-tracker/issues/220)  and [PR](https://github.com/clevercanary/hca-atlas-tracker/pull/228) where the CAP ID is added to the patch request for editing

#### DOD
1. CAP ID is not required
2. CAP ID, if present, is validated to be like  https://celltype.info/project/259/dataset/412 (I don't think we can validate that there are exactly three numbers after the project or after the dataset, but just some numbers.
3. The DOI, title, HCA ID and CELLxGENE ID remain not editable.

frano-m:
Closes #255.

image

### DIFF
--- a/app/components/Detail/components/TrackerForm/components/Section/components/SourceDataset/common/constants.ts
+++ b/app/components/Detail/components/TrackerForm/components/Section/components/SourceDataset/common/constants.ts
@@ -1,4 +1,8 @@
 export const DEFAULT_INPUT_PROPS = {
+  CAP_ID: {
+    isFullWidth: true,
+    label: "CAP ID",
+  },
   CELLXGENE_COLLECTION_ID: {
     isFullWidth: true,
     label: "CELLxGENE collection ID",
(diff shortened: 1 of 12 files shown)


NoopDog (approved):
Super Cool thanks!

PR #325, opened 2024-06-20
feat: update breadcrumbs to match new mocks (#272)

### LINKED ISSUE #272: Update breadcrumbs to match new mocks
#### Need
Currently its not possible to tell the type of entity you are looking at. Source datasets for example don't say "Source Dataset" on them anywhere. Also getting back to the previous page is a bit difficult. To address this the mocks for the breadcrumbs have been updated.

image

See [Figma](https://www.figma.com/design/g60pHZLz78wfl0PvbksMb5/HCA-Tracker?node-id=54-3171&t=qnu5dlshftAqYeEj-1).

### Approach 
Update the breadcrumbs to match the mocks.

#### Questoins!
1. Is the header component configurable per app app? Can we leave the other apps alone for now or is it easier to change all at once and have a single header component?

2. Where does the back button go? I think it's the same as pressing the browser's back button unless you came to the page directly. Then it's to the next level up in the breadcrumb.

frano-m:
Closes #272.

### DIFF
--- a/app/components/Detail/components/TrackerForm/components/Breadcrumbs/common/constants.ts
+++ b/app/components/Detail/components/TrackerForm/components/Breadcrumbs/common/constants.ts
@@ -1,0 +1,53 @@
+import { ROUTE } from "../../../../../../../routes/constants";
+import { Breadcrumb } from "../breadcrumbs";
+
+export const BREADCRUMB_ATLAS: Breadcrumb = {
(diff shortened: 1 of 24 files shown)


frano-m:
image

NoopDog (approved):
Breadcrumbs are never easy. Thanks for working through this @frano-m 🚀

PR #329, opened 2024-06-20
feat: add todo fields to source dataset ui (#327)

### LINKED ISSUE #327: Add TODO fields to source dataset UI
#### Need

The API has been updated to provide the missing fields on the source dataset lists and the component atlas linked source datasets. 

Also add a column for the explore link.  This will currently always be cellxgene but for non cellxgene source dateasets it wont but we can address that on a different ticket. 

Image

Image

Image

frano-m:
Closes #327.


NoopDog (approved):
⭐ ⭐ ⭐ ⭐

Issue #308, opened 2024-06-16
Strip HTML form crossref titles


NoopDog:
#### Need
Some crossref titles have HTML in them like:

<img width="237" alt="image" src="https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/d2eaefb3-6873-48b0-a639-502a1b064ec9">

[10.1101/2021.09.24.460132](https://doi.org/10.1101%2F2021.09.24.460132)
[10.1164/rccm.202305-0924oc](https://doi.org/10.1164%2Frccm.202305-0924oc)

How can we remove the HTML?

The italics can be removed completely.
What can we do with the <sup ? Just remove the HTML but keep the inner content?

NoopDog:
Complete

PR #331, opened 2024-06-20
feat: strip html from fields when getting crossref publication (#308)

hunterckx:
Uses [string-strip-html](https://www.npmjs.com/package/string-strip-html), which is installed as part of this PR


hunterckx:
Tests are currently failing due to string-strip-html being ESM

NoopDog (approved):
Clean! Thx

PR #335, opened 2024-06-21
feat: update ordering on task saved filters (#333)

### LINKED ISSUE #333: Update ordering on task saved filters
### Need
To help present the tasks on a coherent plan, we can order the tasks by additional fields besides the target completion date ascending on the task saved filters.

### Approach

For all saved filters

Image

We currently order by Target Completion ascending. Instead order by:

1. Target Completion ascending then by
2. Atlas ascending
3. Then by status

Once this is in place we can do a separate ticket to custom order the status or do the fake ordering/mapping we did for completed at.

frano-m:
Closes #333.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "hca-atlas-tracker",
       "version": "0.1.0",
       "dependencies": {
-        "@databiosphere/findable-ui": "4.0.0",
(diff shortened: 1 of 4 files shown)


NoopDog (approved):
MIce 🐭 🐭 🐭

Issue #348, opened 2024-06-29
Allow setting of HCA, CELLxGENE and CAP URLs on Source Studies with no DOI - API


NoopDog:
#### Need
Some public source studies are not published in a journal or in BioArxiv for example, see:

https://cellxgene.cziscience.com/collections/ced320a1-29f3-47c1-a735-513c7084d508
https://explore.data.humancellatlas.org/projects/f0f89c14-7460-4bab-9d42-22228a91f185

In these cases the search by DOI will never match in HCA and CxG (and CAP has no API). To handle this we need to allow anyone who can create or edit a source study to also be able to add the links for source studies that have no DOI.

#### Approach
Modify the edit source study API to accept IDs for CAP, CELLxGENE and HCA.
Error if CELLxGENE or HCA ids are set for source studies with a DOI.
Allow CAP ID  to be set regardless of DOI status.

NoopDog:
Merged! Thx

PR #353, opened 2024-07-03
feat: add fields for external ids on unpublished source studies to api (#348)

hunterckx:


app/views/SourceStudyView/common/constants.ts
```diff
@@ -15,12 +15,11 @@ export const FIELD_NAME = {
...
-  FIELD_NAME.CELLXGENE_COLLECTION_ID,
-  FIELD_NAME.HCA_PROJECT_ID,
 ];
 
 export const UNPUBLISHED_FIELDS: SourceStudyEditDataKeys[] = [
   ...NEW_SOURCE_STUDY_UNPUBLISHED_FIELDS,
+  FIELD_NAME.CAP_ID,
   FIELD_NAME.CELLXGENE_COLLECTION_ID,
   FIELD_NAME.HCA_PROJECT_ID,
 ];
```
hunterckx, line 25: I've changed some form-related things to make it so the new fields get sent; as far as I know the only side effect of this is that HCA ID and CELLxGENE ID are no longer used to determine whether the form for a published source study is dirty, which shouldn't cause any issues since those can't be edited anyway

app/views/SourceStudyView/hooks/useEditSourceStudyFormManager.ts
```diff
@@ -122,11 +122,10 @@ function onSuccess(
...
-    fieldKeys.push(...PUBLISHED_FIELDS);
-  }
-  return fieldKeys;
+  const [keysForThisStatus, keysForOtherStatus] =
+    payload.publicationStatus === PUBLICATION_STATUS.PUBLISHED
+      ? [PUBLISHED_FIELDS, UNPUBLISHED_FIELDS]
+      : [UNPUBLISHED_FIELDS, PUBLISHED_FIELDS];
+
+  return keysForOtherStatus.filter((key) => !keysForThisStatus.includes(key));
 }
```
hunterckx, line 131: Also changed this to make it so the fields for the selected publication status are *not* unregistered

frano-m (approved):
LGMT on the FE. Thanks @hunterckx.

NoopDog (approved):
LGTM! Thx

PR #357, opened 2024-07-04
feat: view task details (#321)

### LINKED ISSUE #321: View task details
### Need 
On the tasks list the rows are getting crowded and there could be a bit of information to display. We also want to be able to quickly switch the task detail we are viewing.

### Approach
Implement a drawer with task details, Remove as much info as possible from the table row.

Image

#### Implement  a drawer

[Figma](https://www.figma.com/design/g60pHZLz78wfl0PvbksMb5/HCA-Tracker?node-id=886-22427&t=Qlqkj3a1SsWTqyDw-1). It would open on the right side, be persistent (only closable with “X”), and update its content when another row in the table is clicked. There will likely be future cases where extra descriptions are needed for other tasks. We’re not aiming for a project management tool, but we’re creating one anyway.

#### Reduce / reorder the columns in the table 

1. Task - remove the expected / actual and the contamination of the system name e.g. "in CELLxGENE"
2. System
3. Atlas
4. Network (rename from Bionetwork)
5. PublicationStatus
6. Target Completion
7. Status
8. Completed At

frano-m:
Closes #321.


NoopDog (approved):
LGTM! 🚀

Issue #350, opened 2024-06-30
Add additional columns to the atlas's component atlas list


NoopDog:
### Need

The component atlases list does not summarize the content of the component atlases it currently only lists the title.

<img width="1058" alt="image" src="https://github.com/clevercanary/hca-atlas-tracker/assets/3239697/b1ab9404-dc2c-40ce-9de5-342f950ed438">

We need to add a summary of the contents of the component atlas. Lets dd

1. count of component atlases
2. Rollup of each of the fields on the component atlas's source datasets: Assay, Suspension Type, Tissue, Disease, Cell count

### Approach

1. Add to API, calculating the aggregate values on the server. These can be calculated and saved on the component atlas whenever a source dataset is added or removed from it.
 
2. Update the columns in the UI.

NoopDog:
Complete

PR #364, opened 2024-07-07
feat: add component atlas fields aggregated from source datasets (#350)

hunterckx:
A component atlas's properties are updated when:
- The new migration is run
- A source dataset is added to it
- A source dataset is removed from it
- One of its CELLxGENE source datasets is updated after a refresh


frano-m (approved):
FE LGTM! Thanks @hunterckx.

Issue #367, opened 2024-07-08
Integration Leads can edit a source study and add / edit / delete source datasets and of their own atlases


frano-m:
An integration lead will need full editing access to their own atlases' source studies.

They should be able to:

- [x] Add a source study
- [x] Edit an existing source study e.g. edit "No DOI" fields, or "CAP ID".
- [x] Delete an existing source study
- [x] Add a CELLxGENE collection to a "No DOI" source study
- [x] Remove a CELLxGENE collection from a "No DOI" source study

frano-m:
Resolved by #354.

PR #368, opened 2024-07-08
feat: allow integration leads to use the source study edit api for their atlases (#367)

hunterckx:


NoopDog (approved):
Super, thx! @hunterckx

Issue #370, opened 2024-07-09
Integration Leads can add / delete a non-CELLxGENE source dataset to their own atlas' source study


frano-m:
<img width="1479" alt="image" src="https://github.com/clevercanary/hca-atlas-tracker/assets/18710366/6d5debd0-cda9-4cab-8396-bd3e4df3bc84">

frano-m:
Resolved by #354.

PR #371, opened 2024-07-09
feat: allow integration leads to use source dataset apis for their atlases (#370)

hunterckx:
In addition to allowing creating and deleting source datasets as specified in the ticket, also allows editing source datasets (i.e. editing their titles)


NoopDog (approved):
LGTM, THX

Issue #369, opened 2024-07-08
CELLxGENE admin and content admin can upload a list of DOIs to mark as "in-progress" for CELLXGENE


NoopDog:
#### Need
CELLxGENE does not share private collection ids. We would need to get the individual authors to share the private collection ids associated with CELLxGENE WIP. Will we work out if this is possible we would like to be able to upload a list of DOIs and have tracker mark these DOIs as in progress for CELLxGENE ingest.

### Approach

API /cellxgene/ingest-wip (or something better?)

#### Input:

List of DOIs 

#### Action:

If the source study ingest task for CELLxGENE corresponding to the DOI is found and is TODO, set it to IN-PROGESS; otherwise, do nothing for the given DOI.

#### Response:

List of DOIs with status updated source study ingest task status
List of not-found DOIs
List of DOIs that were already marked TODO
List of DOIs that were already marked DONE

#### Access control
Can be done by CELLXGENE_ADMIN (new role, over all CELLxGENE resources that also has STAKEHOLDER ), and CONTENT_ADMINs

TODO:

1. Decide on URL
2. Decide on request shape
3. Decide on response shape
4. Map allowable task status transitions given a validation status
5. Admin page to hold the form and the response

NoopDog:
complete

PR #374, opened 2024-07-10
feat: add api to cellxgene tasks to in progress (#369)

hunterckx:
- API route is `/api/tasks/cellxgene-in-progress` (PATCH)
- New role CELLXGENE_ADMIN is equivalent to STAKEHOLDER except that it can use the cellxgene-in-progress API
- There is a somewhat-messy form at `/cellxgene-in-progress`


app/apis/catalog/hca-atlas-tracker/common/entities.ts
```diff
@@ -163,6 +163,12 @@ export interface HCAAtlasTrackerActiveUser {
   roleAssociatedResourceIds: string[];
 }
 
+export interface TaskStatusesUpdatedByDOIResult {
+  notFound: string[];
+  notUpdated: Record<TASK_STATUS, string[]>;
+  updated: string[];
+}
```
hunterckx, line 170: This is the format for the API response -- `notFound` contains DOIs not found in the Atlas Tracker; `notUpdated` contains, for each task status, an array of DOIs that were found to have that status and were not updated (e.g. notUpdated.IN_PROGRESS will contain DOIs with tasks that weren't updated because they were already in progress); and `updated` contains DOIs that had their tasks' statuses updated to IN_PROGRESS

(The returned DOIs are normalized and so may not match the input DOIs)

NoopDog (approved):
Super! Thx @hunterckx

PR #375, opened 2024-07-11
feat: make it easier to switch between top level tabs (atlases/tasks) (#340)

### LINKED ISSUE #340: Make it easier to switch between top level tabs (atlases/tasks)
#### Need
It takes several clicks to move between atlases and tasks. We can make this easier by moving the links to the different sections to the top nav.

#### Approach

1. Make it configurable in findableUI if the tab names show in the explore view.
2. Move the tabs to the top nav for the tracker.

Let's discuss the above to see what issues there may be.

frano-m:
Closes #340.


NoopDog (approved):
🚀 Thank you! @frano-m

Issue #376, opened 2024-07-19
Create script to add integration leads to the users table with thier atlases as role related resources


NoopDog:
#### Need
We need to authorize all of the integration leads to access the system, give them the integration lead role, and mark the atlases they are leads for in the database.

#### Approach

Create a script to go over all of the integration leads for each atlas and:

1. Add them to the user table if they are not already there with the integration lead role
2. Add any atlas they are the integration lead for to the role-related resources
3. If the user is already added at the same email address just check if the atlases need to be added.

Let's run this script on an "Import Integration Leads" button on an "admin page". Only let this feature be run by content admins.

NoopDog:
Complete

PR #380, opened 2024-07-22
feat: add api and page to update integration lead users from atlases (#376)

hunterckx:
New API is PATCH `/api/users/integration-leads-from-atlases`, useable by content admins

New page `/integration-leads-from-atlases` allows calling the API


NoopDog (approved):
Super Awesome.. Thx @hunterckx

Issue #377, opened 2024-07-20
Change name of "Tasks" tab to "Reports"


NoopDog:
<img width="975" alt="image" src="https://github.com/user-attachments/assets/1c25db42-f68e-4ed5-b9a4-4fc756618211">

On the UI labels only (and the URL) but not any component names, tables, or big renaming refactor....

Update the 
1. URL to /reports
2. Header Menu item from "Tasks" to "Reports" 
3. Heading on the page from "Tasks" to "Reports"
4. Rename "Saved Filters" to "Saved Reports"

So like:

![Image](https://github.com/user-attachments/assets/3cb59b5d-35c6-4778-a281-729d4afceedc)

NoopDog:
Complete.. will create a new ticket for "saved filters"

PR #381, opened 2024-07-22
feat: rename tasks list to reports (#377)

hunterckx:
It doesn't seem to be possible to change the label of the saved filters


NoopDog (approved):
Ok then!

Issue #379, opened 2024-07-20
Add filter for "Atlas Version" on the atlas and reports (tasks) tab. 


NoopDog:
Add a filter for Atlas Version e.g. 1.0 2.0 in on both the Atlases and Tasks (Reports) tab.
Group it with the Atlas group on the Reports tab and in alpha order on the Atlases tab.

NoopDog:
Complete

PR #383, opened 2024-07-22
feat: add filters for atlas version (#379)

hunterckx:


site-config/hca-atlas-tracker/local/index/tasks/common/columns.ts
hunterckx, line 1: I'm not sure what the difference between these two column lists is -- do both need to contain the new column?

NoopDog (approved):
Super, thx!

Issue #386, opened 2024-07-23
Indicate CELLxGENE ingest status on Source Studies lists and Source Study detail pages


NoopDog:
### Need
The ingest status for CELLxGENE on the lists and the detail pages are currently taken from the presence or absence of a CELLxGENE collection URI on the source study. We now have an "in-progress" status that can not be determined by the presence or absence of the collection ID.

![image](https://github.com/user-attachments/assets/d6c396ce-04a5-43ba-93c0-01d77e9384aa)

<img width="1355" alt="image" src="https://github.com/user-attachments/assets/881a357f-690a-4431-8f64-c3a76c4b5943">

<img width="1667" alt="image" src="https://github.com/user-attachments/assets/4fd43497-12fe-434d-80da-ee28d0f46916">

### Approach
Currently, the API responses for:

-  `/api/atlases/[id]/source-studies` and
- `/api/atlases/[id]/source-studies/[id]` 

have no task information. 

The source study list and source study detail page will need access to the tasks related to each source study to display the in-progress status. 

We can accomplish this by adding a `tasks` key to the source study responses containing information about the tasks associated with the source study.

The data returned should list all the task information for a source study without duplicating the information already on the source study.

#### Example API response

`/api/atlases/[id]/source-studies`

```
[
    {
        "capId": null,
        "cellxgeneCollectionId": null,
        "contactEmail": null,
        "doi": "10.1038/s41467-022-30934-z",
        "doiStatus": "OK",
        "hcaProjectId": null,
        "id": "e771ddc1-00d3-468b-b6e0-d05d85ff0c78",
        "journal": "Nature Communications",
        "publicationDate": "2022-06-07",
        "referenceAuthor": "Yoon",
        "sourceDatasetCount": 0,
        "tasks:"[
            {
                "commentThreadId": null,
                "createdAt": "2024-05-28T20:15:40.949Z",
                "description": "Ingest source study.",
                "differences": [],
                "id": "11a51a05-8b66-4ea4-bef5-2a29b1a7836a",
                "resolvedAt": null,
                "system": "CELLXGENE",
                "targetCompletion": "2024-12-31T23:59:59.999Z",
                "taskStatus": "TODO",
                "updatedAt": "2024-06-16T05:30:07.094Z",
                "validationId": "SOURCE_STUDY_IN_CELLXGENE",
                "validationStatus": "FAILED",
                "validationType": "INGEST",
            }
            ....

         ]
[...]

PR #389, opened 2024-07-26
feat: derive source study status columns from validations (#386)

hunterckx:
The validations returned with source studies don't have atlas-related information included, which is in line with the example response given in the ticket but does mean there's some non-duplicate information that is not included (namely, what other atlases the source study is in)


NoopDog (approved):
Super! Thx! @hunterckx

Issue #395, opened 2024-07-29
Implement NextAuth for Google and Microsoft


NoopDog:
### Need
Our current authentication solution has a few problems:

1.  We only currently support Google as an authentication provider. However, many of the email addresses that users send to us are Microsoft-backed.
2. It's easy to accidentally log out by refreshing the page.
3. When opening a new tab, the new tab is opened in logged-out mode.

### Approach

1. Implement [NextAuth.js](https://next-auth.js.org/) instead of Google Auth using the Google package directly.
2. Do this in the application instead of the package for now. 
3. The NextAuth will set a secure HTTP cookie from the server that functions as the session cookie rather than only keeping the token in JS memory.

NoopDog:
Complete

NoopDog:
Hi, @hunterckx. We're back..  Can you please rebase this?
Also, can you rename the Google secrets as read in from the environment variables to:

GOOGLE_CLIENT_ID and
GOOGLE_CLIENT_SECRET

NoopDog:
ok looks through once more and we can deploy to dev and test! Thx @hunterckx

PR #398, opened 2024-07-31
feat: use `next-auth` for authentication (#395)

hunterckx:


hunterckx:
Associated Findable UI pull request: databiosphere/findable-ui#168

NoopDog:
Hi @hunterckx, can you rebase this, please?

hunterckx:
Rebased and upgraded `next-auth`, but something I still need to do is ensure that the tests work

(Also I didn't check that `next-auth` is still working after the update because the CELLxGENE datasets API is having trouble which causes problems for running the app, but it's just patch versions involved in the update so it's probably fine?)

hunterckx:
Alright, tests are working now

NoopDog (approved):

Issue #406, opened 2024-08-01
Accept a description field on atlases and integration objects in the API


NoopDog:
#### Need
We want to have a description field to hold text like:

<img width="859" alt="image" src="https://github.com/user-attachments/assets/8bf8d7a1-c688-465c-b78d-51323fb0cd5a">

For atlases and 

<img width="871" alt="image" src="https://github.com/user-attachments/assets/2815d597-8d19-401f-b67c-589b0462bb1c">

For the integration objects. 

#### Approach
On the /atlas and /atlas/component-atlas APIs, add a "description" field in the database and on the entity object with the same permission as the form (CONTENT_ADMINS and INTEGRATION_LEADS for the atlas).

This can be protected with a max length validation. 

This can be just a string field, and it will be interpreted as markdown, with the markdown processing done on the client.

NoopDog:
Complete

PR #409, opened 2024-08-03
feat: add description fields to atlases and integration objects (#406)

hunterckx:
Create and edit APIs for atlases and integration objects accept an optional `description` field, which is limited to 10000 characters by the validator


NoopDog (approved):
🚀

Issue #420, opened 2024-09-21
Address non-findable-ui related security issues and dependabot version bumps


NoopDog:
- For the packages that are coming from findable-ui peer dependencies, fix any version bumps or security warnings.

NoopDog:
Complete thx!

PR #421, opened 2024-09-22
chore: update packages to resolve vulnerabilities (#420)

hunterckx:
Running `npm audit fix` resolved most of what NPM and Dependabot reported; it apparently left an outdated version of `ws` though, which was a dependency of a dependency of Storybook. To my knowledge we haven't actually used Storybook for this app, so I've uninstalled it and the outdated `ws` package is now gone from `package-lock.json`.


NoopDog (approved):
Super thanks @hunterckx

Issue #419, opened 2024-09-13
Add validation for the HCA Data Repository for network and atlas


NoopDog:
### Need 
The HCA Data Explorer shows the atlases and BioNetworks that are associated with a given study. The HCA data wranglers require a report to indicate the atlases and BioNetworks that are missing from projects in the HCA Data repository.

### Approach 

1. Add a new HCA Data Repository validation called "Link project to HCA BioNeworks and Atlases."
2. Validate each source study's project. 
3. Make the validation depends on if the project exists in the HCA Data Repository (skip the validation until the project exists)
4. Validate the project on save of a source study or when the projects refresh normally happens.
5. Crate a saved filter, "HCA Data Repository - Missing Network or Atlas that filters on todo instances of this event. 

#### Validating a project's networks in the HCA Data Repository
Every project corresponding to a tracker's source study should have one or more atlases listed in the Azul response in the project's `bionetworkName` field for example:

```
 "bionetworkName": [
    "Genetic Diversity",
    "Immune"
 ],
```

#### Validating a projects atlas's are listed in the HCA Data Repository
Every project associated with  a source study should have one or more atlases listed in the projects Azul response like:

```
"tissueAtlas": [
                        {
                            "atlas": Lung,
                            "version": "v1.0"
                        },
                        {
                            "atlas": "Blood",
                            "version": "v1.0"
                        }
                    ],

```

#### Description
For the expected and actual in the description, make a sentence like.

Missing Networks: Lung
Missing Atlases Lung v1.0, Lung v2.0, Brain v1.0

NoopDog:
Complete

PR #422, opened 2024-09-26
feat: add validation for hca project networks and atlases (#419)

hunterckx:
- Source studies provided to validations now have some atlas properties attached
- Differences listed in validation results can now be arrays, in which case, in the app, the values from `expected` that are missing from `actual` are displayed
- Networks are compared case-insensitively, since they're stored in lowercase in the tracker but appear to be capitalized in the data browser


NoopDog (approved):
OK great! Thx @hunterckx

Issue #423, opened 2024-09-28
Add form for creating/editing a user


hunterckx:
- Only usable by content admins
- Allow creating integrations leads with a multiselect for associated atlases

NoopDog:
Complete! Thx

PR #424, opened 2024-10-02
feat: add form to create a user (#423)

hunterckx:
- New API route `/api/users/[id]`: GET to get user, PATCH to update user (using same payload as existing create API)
  - In the `API` enum, `/api/me` is renamed from `USER` to `ACTIVE_USER` so that `USER` can be used for the new API
  - Similarly, `useFetchUser` is renamed to `useFetchActiveUser`
- New forms (using the same underlying component) for creating and editing users
- The "Add Atlas" button is generalized to a `HeroButton` component, which is used by both the atlases hero and the users hero
- The forms have API mappers to:
  - Convert the enabled/disabled select value to a boolean
  - Remove associated atlases for roles other than integration lead

### DIFF
--- a/__tests__/api-users-create.test.ts
+++ b/__tests__/api-users-create.test.ts
@@ -1,6 +1,5 @@
 import { NextApiRequest, NextApiResponse } from "next";
 import httpMocks from "node-mocks-http";
-import { testApiRole } from "testing/utils";
 import {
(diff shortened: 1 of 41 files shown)


frano-m (approved):
Niiiice @hunterckx 🚀

Issue #426, opened 2024-10-07
Make enabled/disabled status affect user permissions


hunterckx:

NoopDog:
Complete

PR #428, opened 2024-10-13
feat: prevent users with disabled accounts from accessing the app (#426)

hunterckx:
- APIs that require a registered account now also require an enabled account
  - Many of the changed files are tests updated to check this
- A message is shown when a user tries to log in with a disabled account


NoopDog (approved):
Thanks @hunterckx !

Issue #429, opened 2024-10-15
Trim email input before validating


NoopDog:
<img width="512" alt="image" src="https://github.com/user-attachments/assets/0041919d-90eb-40e3-904e-8eb16966e95e">

On the user form, when entering an email, copying and pasting in an email may result in leading or trailing white space in the input data.
Currely, if this happens, saving will fail validation on the email.

Instead of failing validation, trim the input value in place before validation for the user so they don't have to manually delete any leading/trailing whitespace.

NoopDog:
Complete thx!

PR #430, opened 2024-10-16
feat: trim whitespace from email in user forms (#429)

hunterckx:
I've removed the "strict" setting from the forms' schemas, which allows the email to be automatically trimmed before being sent to the API; as far as I know removing that setting won't have any other effects -- I believe it's a carryover from the API schemas, where we wanted to prevent values of the wrong type from being cast to the correct type and thus accepted

(Note: I've also tacked on a change to the user create handler, because I came across an issue where a validation error that happened server-side didn't get returned to the client)


frano-m (approved):
Nice @hunterckx ! Thank you!

Issue #432, opened 2024-10-22
Add "Validating an Atlases Source Study List" as a doc to the help and docs section


NoopDog:
The google doc is here
https://docs.google.com/document/d/1a8jqmF5T9awAT55t3ZbTsajLBoKmnOjwkb0s-7vI5XA/edit?tab=t.0#heading=h.ll0jzrtdyuy7

1. remove the other empty guide pages.
2. Add this as the first real guide. Feel free to fix any text or re-take the screen shots.

NoopDog:
Complete

PR #436, opened 2024-10-23
feat: add "validating an atlas's source study list" page (#432)

hunterckx:


frano-m (approved):
Niiice @hunterckx thank you!

Issue #467, opened 2024-12-02
Allow linking of source datasets to atlases


NoopDog:
### Need
Currently, source studies are linked to atlases, and source studies are linked to source datasets. This allows source datasets for an atlas to be accessed through source studies. However:

1. Not all source datasets in a source study are necessarily associated with the atlas. 
2. We want to be explicit about which source datasets are included in the atlas.

To address this, we will allow datasets to be linked directly to an atlas from a source study making them truly "source datasets".

#### Data Model

Add an array on the atlas model to hold the IDs of linked source datasets near [here] like `source_datasets` (https://github.com/clevercanary/hca-atlas-tracker/blob/main/app/apis/catalog/hca-atlas-tracker/common/entities.ts#L212).

#### Read
Return the list of source datasets associated with the atlas in a GET like /atlas/id/source-datasets API call. 

#### Link
Create an API allowing CONTENT_ADMINS and INTEGRATION_LEADS to link a source dataset to an atlas by sending the Atlas ID and the source dataset ID.  

Q: How can we validate that the dataset's source study is an atlas source study?
  

#### Unlink 
Create an API allowing CONTENT_ADMINS and INTEGRATION_LEADS to unlink a source dataset from an atlas by sending the Atlas ID and the source dataset ID.  

#### User navigates to an Atlases Source Study Dataset list

On the Atlas > Source Study > Dataset page, add a drop-down selector to indicate if each dataset is used or not used by the atlas.

Initialize the content of the tags from the GET call above. If the dataset's ID is not associated with the atlas, then it is not "in use" by that atlas.

<img width="825" alt="image" src="https://github.com/user-attachments/assets/03ab3f95-ce4c-45a9-b290-e392b05d7b7a">

#### User modifies a tag
When the user modifies a "Used in Atlas"  tag, the API is called to link/unlink the dataset to the atlas, sending the atlas ID and the dataset ID to the link or unlink endpoint.

NoopDog:
Complete! Thx

PR #479, opened 2024-12-10
feat: allow linking source datasets to atlases (#467)

hunterckx:
- Checks are done for:
  - Whether a source dataset can be linked to an atlas (is the dataset's study a source study of the atlas?)
  - Whether a source study can be deleted from an atlas (does it have no source studies linked to the atlas?)
  - Whether a source dataset can be deleted (is it linked to no atlases?)
- When a CELLxGENE source dataset is automatically deleted, it's removed from any atlases it's linked to
- Changing a dropdown calls the API immediately to link/unlink the dataset (this is what's described in the ticket but I believe an alternative idea was brought up at some point)
- Not 100% certain my `useSetLinkedAtlasSourceDatasets` hook is reliable -- I modeled it roughly on how unlinking datasets from component atlases works, but I didn't use `useFetchDataState` and I'm not sure what difference that makes
- Getting the counts of linked datasets is done in a slightly convoluted way client-side (see `viewSourceStudies.tsx`) -- not sure if that should be done more server-side or what


app/components/Detail/components/ViewAtlasSourceDatasets/viewAtlasSourceDatasets.tsx
```diff
@@ -0,0 +1,57 @@
...
+        {atlasSourceDatasets.length > 0 && (
+          <Table
+            columns={getAtlasSourceDatasetsTableColumns()}
+            gridTemplateColumns="max-content minmax(200px, 1fr) minmax(180px, auto) repeat(4, minmax(88px, 0.4fr)) auto"
+            items={atlasSourceDatasets.sort(sortSourceDataset)}
+            tableOptions={TABLE_OPTIONS}
+          />
+        )}
+        <TablePlaceholder
+          canEdit={canEdit}
```
frano-m, line 37: Nice! The placeholder might have been created for a different use-case i.e. when a user has view access, but not editing access ... Would you mind updating the `TablePlaceholder` component to take a single prop like `showPlaceholder` and update each use of this component accordingly.

frano-m (approved):
Niiice @hunterckx thank you!

NoopDog (approved):

Issue #505, opened 2024-12-23
Add CELLxGENE collection title to the Atlas detail collection link input when available


NoopDog:
#### Need
We can enter the  link to the Atlas in CELLxGENE on the atlas detail page. However, it's not clear without clicking through which Collection it is.  
<img width="1346" alt="image" src="https://github.com/user-attachments/assets/2e632704-4103-4a04-814f-cde1354f29c9" />

Lets show the collection title below the link input in the same style we display the publication string for a DOI. See the example above. The collection name might be longer so hopefully that can wrap in place if needed.

NoopDog:
Complete

PR #506, opened 2024-12-24
feat: add title to atlas cellxgene collection input (#505)

hunterckx:
I've set it up so the collection title is retrieved dynamically from the CELLxGENE data cache on the backend, rather than being stored in the database; because of this, converting a database atlas to an API atlas requires imports that only work on the backend, so I've created a new `backend-utils` file


NoopDog (approved):
Thanks @hunterckx !

PR #535, opened 2025-01-10
feat: add download buttons to atlas source datasets (#526)

### LINKED ISSUE #526: User can download CELLxGENE source datasets
#### Need

For source datasets, we need to allow users to download the dataset so that it can be inspected locally or uploaded into a different system, e.g. CAP.

#### Approach
image

On the datasets page add a "Download" icon link
image

Similar to https://data.humancellatlas.org/hca-bio-networks/lung/atlases/lung-v1-0

Have the link directly download the file (instead of having a popup like on the link above)

Lets discuss if there is any issue getting / building the download link.

hunterckx:
Closes #526

- URL are constructed under the assumption that for any dataset version ID there will be a file with a URL of the form `https://datasets.cellxgene.cziscience.com/${versionId}.h5ad` (this appears to be the case for all current datasets on CELLxGENE)
- Only the atlas source datasets list has the download button


NoopDog (approved):
Super thx!

PR #536, opened 2025-01-10
feat: allow setting past target completions (#534)

### LINKED ISSUE #534: Allow target completion on atlases to be set in the past
Currently only future quarters can be assigned and some have completed in the past.
Also (probably related) on the edit form show the current selected value in the drop down

Only shows past dates...

image

Completed one does not sow the target date on the form as it is in the past.

image

hunterckx:
Closes #534


NoopDog (approved):
NIce!

PR #541, opened 2025-01-11
feat: add additional grouping options for tasks and remove default grouping (#538)

### LINKED ISSUE #538: Add additional "Group By" columns for the "Reports" tab
#### Need
Currently, the only groupable column is "Target Completion."

Allow Grouping by everything except

1. Created at 
2. Updated at
3. Completed at

If the multi-valued fields are an issue, we can skip them for now. 

Update:

1. Let's skip the multi-valued fields.
2. Please remove the default grouping on "Target Completion". Initially the page should display ungrouped. (unless its part of a saved filter, of course)

hunterckx:
Closes #538

In addition to multi-valued fields and dates, I've left out entity type, since we only have validations for source studies


NoopDog (approved):
Too easy :)

PR #543, opened 2025-01-14
feat: show per-system ingestion task counts in atlas list rather than all tasks (#539)

### LINKED ISSUE #539: Update atlases page to show tasks per component
#### Need 
On the Atlas tab, the current columns are:

image

To give a more complete status of the project from the top level, it would be helpful to list the ingest-related tasks for each component ( task type ingest Source Study.)

1. CELLxGENE 23/44
2. CAP  0/44
3. Data Repo 33/44

#### Approach
Let's investigate if this can be done similarly to how the current "tasks" count is generated.

hunterckx:
Closes #539


app/apis/catalog/hca-atlas-tracker/common/entities.ts
```diff
@@ -510,6 +520,13 @@ export enum VALIDATION_TYPE {
   METADATA = "METADATA",
 }
 
+export enum VALIDATION_DESCRIPTION {
```
hunterckx, line 523: Changed `VALIDATION_DESCRIPTION` to an enum so it could be used for a parameter of `getTaskCountUrlObject`

migrations/1736802716583_atlas-validation-counts-by-system.ts
```diff
@@ -0,0 +1,19 @@
...
+  CAP: { completedCount: 0, count: 0 },
+  CELLXGENE: { completedCount: 0, count: 0 },
+  HCA_DATA_REPOSITORY: { completedCount: 0, count: 0 },
+};
+
+export function up(pgm: MigrationBuilder): void {
+  pgm.sql(
+    `UPDATE hat.atlases SET overview=overview||'{"ingestionTaskCounts": ${JSON.stringify(
+      DEFAULT_INGESTION_COUNTS
+    )}}'`
```
hunterckx, line 13: Wasn't totally sure about whether it's acceptable to use a template literal here, but in theory it should be fine since it's inserting a constant

testing/constants.ts
```diff
@@ -988,6 +988,7 @@ export const SOURCE_STUDY_PUBLISHED_WITH_NO_HCA_OR_CELLXGENE: TestPublishedSourc
 
 export const SOURCE_STUDY_PUBLISHED_WITH_CAP_AND_NO_CELLXGENE: TestPublishedSourceStudy =
   {
+    capId: "cap-id-published-with-cap-and-no-cellxgene",
```
hunterckx, line 991: Discovered one of the test entities wasn't testing what it was supposed to, so fixed that

NoopDog (approved):
👍

PR #544, opened 2025-01-14
feat: remove cellxgene link to make room in atlas source datasets table (#542)

### LINKED ISSUE #542: Remove Explore in cellxgene link from source datasets
The row has too much info in it and it is exceeding the max width.
image

Until we can get a design and update this, let's remove the explore link and give width to the dataset, publication and source study. Well this is probably only enough to let the cell count show. See what you can do.

hunterckx:
Closes #542


NoopDog (approved):
Thx

Issue #547, opened 2025-01-15
Add Metadata Spreadsheet (Golden Sheet) link to the Atlas list


NoopDog:
- Show as just  a link like "metadata spec" no need to put the full URL.
- Open in a new tab
- If no sheet leave the cell blank

jpaten:
- Insert after "Biological Network"
- Text should always be "Metadata Spec" when present, link should vary
- Link should open in a new tab
- Column should be enabled by default

jpaten:
Make column not sortable

PR #552, opened 2025-01-17
feat: added metadata specification column (#547)

jpaten:
- Added metadata specification column
- Displays a blank value when the value of `metadataSpecificationUrl` is `null`, otherwise displays "Metadata Spec" and links to the url
- The url opens in a new tab


NoopDog:
@jpaten  sb I change the requirements plz. Instead of the text "metadata spec" in the cell can you show the truncated url. There is an example of this now on the dataset list showing the link to the dataset metadata entry sheet.

NoopDog:
See this pr here for an https://github.com/clevercanary/hca-atlas-tracker/pull/559

hunterckx (approved):
Thanks @jpaten!

PR #559, opened 2025-01-19
feat: add metadata spreadsheet link to atlas source dataset list (#549)

### LINKED ISSUE #549: Add metadata spreadsheet link to source datasets list
#### Need

- Add after the "Source Dataset"  column
- The column header is "Metadata Entry Sheet"
- truncate the URL so it won't fold
- add an external link icon if it's easy.
- Open in a new tab
- Show blank in the cell if there is no metadata spreadsheet

hunterckx:
Icon not included -- I don't think it would be straightforward to add

Closes #549


NoopDog (approved):
Thanks!

PR #574, opened 2025-01-31
feat: add related entity urls to all validations and show "N/A" when unavailable (#572)

### LINKED ISSUE #572: Show resource link on all detail views on the report list
#### Need
Not all task types list the "Resource" on the flyout for example link to project to HCA Bionetwork does...

Image

But ingest source study does not.  Likely, this is because ingest source study only has the resource if the task is complete.

Image

Looks like ingest source study is the only one where it is missing.

#### Approach
Add the resource section for the flyout and use N/A if it is not available but always show the section. Link to HCA, CELLxGENE or CAP as appropriate

hunterckx:
Closes #572


NoopDog (approved):
Clean! thx!

PR #586, opened 2025-02-20
feat: update name and content of source study datasets column for more clarity (#585)

### LINKED ISSUE #585: On the Source Studies tab rename "Datasets" to "Datasets Used"
#### Need

There is some confusion with the Datasets column.

Image

#### Request

1. Update the heading to "Datasets Used"
2. When there are no datasets, use "--" instead of "0/0"

This won't totally fix the problem that this field is not understood, but hints in the right directory.

hunterckx:
Closes #585


NoopDog (approved):
LGTM! Thx @hunterckx

Issue #589, opened 2025-02-22
Indicate HCA Data Repository primary data ingest status on the source studies tab


NoopDog:
#### Need
The HCA Data Repository currently shows a green check if the source study DOI can be found on an HCA Repo project. The HCA Data repository is meant to hold the primary sequencing data (FASTQ) used to generate matrices used in the atlases. 

Sometimes, the source study is ingested into the HCA repo, but only the original author matrices are ingested and not the primary data. We need to indicate that these projects with no primary data are not complete. 

We already have the primary data states available from the reports tab and need to integrate that with the source studies' request data so we can display the information like:

<img width="720" alt="Image" src="https://github.com/user-attachments/assets/f93de1a8-3427-442c-99d8-dde718bc755b" />

MillenniumFalconMechanic:
Completed.

PR #593, opened 2025-02-25
feat: incorporate primary data status into source study hca status (#589)

hunterckx:
Notes on contents:
- I've adjusted a couple of the labels ("To do" and "Complete") to be consistent with task statuses
- The HCA status is set as "Primary data blocked" if the ingest task is `DONE` and the primary data task is`BLOCKED` -- not certain this is exactly how it will end up working, as currently there's no way for the primary data task to have the `BLOCKED` status, but I believe the plan is to in the future allow primary data to be set as blocked
- Despite the mock showing "In progress" as a possibility for HCA status, I didn't include it as a possibility for HCA here, since to my knowledge there's no current or planned way of setting the HCA ingest task or the primary data task as being in progress

Other considerations:

The colors are not entirely accurate to the mock -- the foreground color for "To do" is noticeably lighter in the mock, and I've used the blue `info` color for "In progress" for now because we don't have a yellow available as far as I know

Here's a screenshot with test values showing all of the HCA statuses from the mock:

Screenshot 2025-02-24 at 10 48 19 PM

---

Additional thought -- I have not pushed this, but I wonder if it would be good to show when CAP is blocked as in the below example, rather than just calling it "To do":

Screenshot 2025-02-25 at 1 19 26 PM

### DIFF
--- a/app/components/common/CustomIcon/components/RequiredIcon/requiredIcon.tsx
+++ b/app/components/common/CustomIcon/components/RequiredIcon/requiredIcon.tsx
@@ -1,19 +1,23 @@
-import { smokeDark } from "@databiosphere/findable-ui/lib/theme/common/palette";
+import { CustomSVGIconProps } from "@databiosphere/findable-ui/lib/components/common/CustomIcon/common/entities";
 import { SvgIcon } from "@mui/material";
 
 /**
  * Custom required icon.
  */
 
-export const RequiredIcon = (): JSX.Element => {
+export const RequiredIcon = ({
+  fontSize = "small",
+  viewBox = "0 0 20 20",
+  ...props /* Spread props to allow for Mui SvgIconProps specific prop overrides e.g. "htmlColor". */
+}: CustomSVGIconProps): JSX.Element => {
   return (
-    <SvgIcon viewBox="0 0 20 20" fontSize="small">
+    <SvgIcon viewBox={viewBox} fontSize={fontSize} {...props}>
       <circle
         cx="10"
         cy="10"
         r="8"
         fill="transparent"
-        stroke={smokeDark}
+        stroke="currentColor"
         strokeWidth="2"
       />
     </SvgIcon>
--- a/app/components/common/CustomIcon/components/RequiredIcon/requiredIcon.tsx
+++ b/app/components/common/CustomIcon/components/RequiredIcon/requiredIcon.tsx
@@ -1,5 +1,4 @@
-import { CustomSVGIconProps } from "@databiosphere/findable-ui/lib/components/common/CustomIcon/common/entities";
-import { SvgIcon } from "@mui/material";
+import { SvgIcon, SvgIconProps } from "@mui/material";
 
 /**
  * Custom required icon.
@@ -9,7 +8,7 @@ export const RequiredIcon = ({
   fontSize = "small",
   viewBox = "0 0 20 20",
   ...props /* Spread props to allow for Mui SvgIconProps specific prop overrides e.g. "htmlColor". */
-}: CustomSVGIconProps): JSX.Element => {
+}: SvgIconProps): JSX.Element => {
   return (
     <SvgIcon viewBox={viewBox} fontSize={fontSize} {...props}>
       <circle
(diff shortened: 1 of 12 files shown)


NoopDog:
Logic is all good @hunterckx. We don't have in progress yet for HCA or blocked for anything. 

@frano-m how do we add colors ?

app/components/common/CustomIcon/components/RequiredIcon/requiredIcon.tsx
```diff
@@ -1,19 +1,23 @@
...
 /**
  * Custom required icon.
  */
 
-export const RequiredIcon = (): JSX.Element => {
+export const RequiredIcon = ({
+  fontSize = "small",
+  viewBox = "0 0 20 20",
+  ...props /* Spread props to allow for Mui SvgIconProps specific prop overrides e.g. "htmlColor". */
+}: CustomSVGIconProps): JSX.Element => {
```
frano-m, line 12: I'm trying to move closer to Mui Props wherever possible, would you mind updating line 12 `CustomSVGIconProps` to just `SVGIconProps` from `Mui`.

frano-m:
@NoopDog and @hunterckx, regarding the Chip colours:

We can actually extend Mui's palette with `caution` similar to the way we do this for `smoke` in [Findable](https://github.com/DataBiosphere/findable-ui/blob/849e5cf5898f71be210436863c9b10baf85a8427/types/data-explorer-ui.d.ts#L125-L145) and [here](https://github.com/DataBiosphere/findable-ui/blob/238f6d3e3fb2ba9737d2daf13739e13d98713f11/src/theme/common/palette.ts#L210-L215).


1. Create a new `theme.d.ts` file in Tracker.
2. Use module augmentation to add the new colour to Theme and ThemeOptions.

```
import type {} from "@mui/material/Button";
import { PaletteColorOptions } from "@mui/material/styles";

/**
 * Palette definitions.
 */
declare module "@mui/material/styles/createPalette" {
  interface Palette {
    caution: PaletteColor;
  }

  interface PaletteOptions {
    caution?: PaletteColorOptions;
  }
}
```

3. Do this for ChipProps as well. See [Findable](https://github.com/DataBiosphere/findable-ui/blob/849e5cf5898f71be210436863c9b10baf85a8427/types/data-explorer-ui.d.ts#L93-L97). This will ensure the status chip now accepts `caution` as a colour.

4. Deep merge the new palette color into the theme.

5. Add styles to the theme for the Mui Chip component (see BRC Analytics for an example). Use below as an example. Note the additional variant for `color: default` and `variant: "status"`, which should target a "status" chip with default colouring. Let me know if you have any problems - I may have missed something out!

```
export const MuiChip = (theme: Theme): Components["MuiChip"] => {
  return {
    styleOverrides: {},
    variants: [
      {
        props: { color: "default", variant: "status" },
        style: {
          backgroundColor: theme.palette.smoke.main,
          color: theme.palette.ink.light,
        },
      },
      {
        props: { color: "caution" },
        style: {
          backgroundColor: theme.palette.caution.light,
          color: theme.palette.caution.main,
        },
      },
    ],
  };
};
```

hunterckx:
Thanks @frano-m -- updated the props type and will see if I can get a new palette color working!

hunterckx:
> We don't have in progress yet for HCA or blocked for anything.

@NoopDog We do have an existing `BLOCKED` status that's used for the CAP ingest task when the source study isn't in CELLxGENE -- hence my idea of updating the CAP column of the source study list to reflect that

NoopDog:
I see @hunterckx that will confuse, I think. So let's leave CAP all TODO on the Source Study list for now.

app/theme/common/components.ts
```diff
@@ -41,3 +41,25 @@ export const MuiButton = (theme: Theme): Components["MuiButton"] => {
     },
   };
 };
+
+export const MuiChip = (theme: Theme): Components<Theme>["MuiChip"] => {
+  return {
+    variants: [
+      ...(theme.components?.MuiChip?.variants ?? []),
```
hunterckx, line 48: Not sure if there's a better way to do this -- I tried calling `deepmerge` with the base theme and a set of options including this MuiChip definition and it caused the other status status badges to have strange colors

NoopDog (approved):
Nice! Thx @hunterckx

PR #653, opened 2025-05-31
fix: accept project urls for cap id rather than dataset urls (#649)

### LINKED ISSUE #649: Can't save CAP Project URL on a Source Study
#### To Reproduce

1. Go to a source study
2. Add a CAP project URL the source study such as https://celltype.info/project/580
3. A validation error displays and the id wont save

Image

What validation are we doing that fails? Are we mistakingly checking this is CAP dataset id like https://celltype.info/project/580/dataset/1265?

hunterckx:
Closes #649

Also updates the backend to restrict CAP URLs by the regex, which was not being done before -- was there a reason for that?


NoopDog (approved):

NoopDog:
Merged! Thx @hunterckx !

NoopDog:
Merged! Thx @hunterckx !

PR #773, opened 2025-08-21
feat: make the title not editable on the integrated object detail page general Info section  #755 (#755)

### LINKED ISSUE #755: Make the title not editable on the Integrated Object detail page General Info section
Add removal of editing, and related form functionality from the Integrated Object page.

frano-m:
Closes #755.

This pull request introduces a minor update to the form controller configuration by adding a new property to the allowed input props and updating the title field to be read-only.

- **Form Input Props Update:**
  - Added `readOnly` to the list of allowed properties in the `PickedInputProps` type in `entities.ts`, enabling form inputs to be set as read-only.

- **Controller Configuration Update:**
  - Set the `readOnly` property to `true` for the `TITLE` field in the controller configuration, making the title input non-editable.

### DIFF
--- a/app/components/common/Form/components/Controllers/common/entities.ts
+++ b/app/components/common/Form/components/Controllers/common/entities.ts
@@ -17,7 +17,9 @@ type PickedInputProps =
   | "label"
   | "helperTextProps"
   | "isFullWidth"
-  | "placeholder";
+  | "placeholder"
+  | "readOnly";
+
 type PickedSelectProps = "displayEmpty" | "label";
 
 export interface ControllerConfig<T extends FieldValues, R = undefined> {
--- a/app/views/ComponentAtlasView/common/controllers.ts
+++ b/app/views/ComponentAtlasView/common/controllers.ts
@@ -12,6 +12,7 @@ const TITLE: CommonControllerConfig = {
   inputProps: {
     isFullWidth: true,
     label: "Title",
+    readOnly: true,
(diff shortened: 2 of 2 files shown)


frano-m:
Hey @hunterckx not sure if we need `.required()` on the `schema` [here](https://github.com/clevercanary/hca-atlas-tracker/pull/773/files#diff-602cc3a778a67102b7b86922562c8f92dae60f3a227cab7e34a681f60410b510R11)?

NoopDog:
Do we need a schema at all here? This is not a form.

hunterckx:
> Do we need a schema at all here? This is not a form.

It is still implemented mostly as a form; hmm....depending on how much work it would be to change that, I wonder if the best thing might be to leave it as a non-editable form for now and change it on another ticket?

> Also title will sometimes empty string

FWIW that's not accurate to how I've implemented it in #766, where I have the title default to the filename (although regardless I don't think it matters whether the title is set to required when it can't be edited)

Other than the need to make this not a form at some point, this looks good to me!

frano-m:
Agree @hunterckx, wondered whether to continue using the forms / controllers, felt the code change would be more than the hassle? Thoughts @hunterckx, @NoopDog ?

NoopDog:
Yes, leave it as a form for now. @hunterckx, looks ok otherwise?

hunterckx (approved):

PR #796, opened 2025-09-03
feat: rename datasets used to datasets on alas source studies (#795)

### LINKED ISSUE #795: Rename  datasets used  to datasets on alas source studies
Image

- Rename the column to "Datasets"
- Show only the lower number (atlas datasets)
- Remove the code if any that did the 3 of 6 type implementation - we wont need it.

Ping with any questions

frano-m:
Closes #795.

This pull request refactors how the source study datasets count column is built and displayed in the HCA Atlas Tracker catalog. The main changes simplify the code by removing an unused builder function, updating how the datasets count link is rendered, and adjusting table column definitions and visibility.

**Table configuration and column definition updates:**

* The `id` column is now included in the table columns and hidden by default in the table options, improving internal row identification while keeping the UI clean. [[1]](diffhunk://#diff-24930035ec067eda637648ac5f14156b9d7920ac216c35cc2c21800155e4531bL9-R9) [[2]](diffhunk://#diff-b40c9cfae34bbe2fe51e79be552b0af2ebf79552105c70236d5978760ebbe81aR1265-R1269)

**Source datasets count column refactor:**

* The `buildSourceStudySourceDatasetCount` function was removed, and its logic was replaced by a simplified inline implementation in the column definition, reducing complexity and improving maintainability. [[1]](diffhunk://#diff-b40c9cfae34bbe2fe51e79be552b0af2ebf79552105c70236d5978760ebbe81aL634-L660) [[2]](diffhunk://#diff-b40c9cfae34bbe2fe51e79be552b0af2ebf79552105c70236d5978760ebbe81aL1773-R1761)
* The datasets count column header was renamed from "Datasets Used" to "Datasets" for clarity.
* The column definition for the source datasets count no longer requires `atlasLinkedDatasetsByStudyId`, streamlining function signatures and usage. [[1]](diffhunk://#diff-b40c9cfae34bbe2fe51e79be552b0af2ebf79552105c70236d5978760ebbe81aR1265-R1269) [[2]](diffhunk://#diff-b40c9cfae34bbe2fe51e79be552b0af2ebf79552105c70236d5978760ebbe81aL1773-R1761)

image

### DIFF
--- a/app/components/Detail/components/ViewSourceStudies/constants.ts
+++ b/app/components/Detail/components/ViewSourceStudies/constants.ts
@@ -6,6 +6,6 @@ export const TABLE_OPTIONS: ListConfig<HCAAtlasTrackerSourceStudy>["tableOptions
   {
     getRowId: (row) => row.id,
     initialState: {
-      columnVisibility: { [COLUMN_IDENTIFIER.ROW_POSITION]: true },
+      columnVisibility: { id: false, [COLUMN_IDENTIFIER.ROW_POSITION]: true },
     },
   };
--- a/app/components/Detail/components/ViewSourceStudies/constants.ts
+++ b/app/components/Detail/components/ViewSourceStudies/constants.ts
@@ -6,6 +6,6 @@ export const TABLE_OPTIONS: ListConfig<HCAAtlasTrackerSourceStudy>["tableOptions
   {
     getRowId: (row) => row.id,
     initialState: {
-      columnVisibility: { id: false, [COLUMN_IDENTIFIER.ROW_POSITION]: true },
+      columnVisibility: { [COLUMN_IDENTIFIER.ROW_POSITION]: true },
     },
   };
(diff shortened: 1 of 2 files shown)


hunterckx:
I've updated this to only count atlas-linked datasets as specified in the ticket, but I want to check, because I also changed the way that the source study ID is accessed in the column config: is there a reason to prefer `row.getValue("id")` over `row.original.id`? I've changed the column to use the latter; `row.original` seems preferable to me, mainly because the type is more reliable (`row.getValue` basically just assumes that it's returning the right type, while `row.original` has the type specified as a parameter to `ColumnDef` and so its properties will have the appropriate types)

hunterckx:
From some brief research, it looks like the difference between `row.getValue` and `row.original` is the the former is the value produced by any processing that the table does (namely via the `accessorFn` if one is specified), while the latter, as suggested by the name, is just the original input data. For the source study ID, I think it's pretty unambiguous that we want the original value, so I believe my change is reasonable

NoopDog (approved):

PR #841, opened 2025-09-15
feat: use metadata from files in component atlas api responses (#837)

### LINKED ISSUE #837: Update integrated object APIs to take metadata from files

hunterckx:
Closes #837

We had a database constraint requiring that the value in the `title` column of `hat.component_atlases` be unique per associated atlas; I've added a migration to drop that constraint, mainly because I wanted to set that title field to empty string for the test component atlases, but presumably it's also useful in general to not have to worry about that constraint now that we're not actually using that column


NoopDog (approved):
LGTM! @hunterckx 🚀 thx

PR #844, opened 2025-09-16
fix: use latest file version where unspecified for source dataset and integrated object apis (#840)

### LINKED ISSUE #840: Only use latest file version when querying files for source dataset and integrated object APIs

hunterckx:
Closes #840

All queries that reference the files table for source dataset and integrated object APIs have been updated, except for the query for getting a specific integrated object, which takes a file ID and so can only ever return that specific file record


NoopDog (approved):
LTGM! Thx @hunterckx

PR #877, opened 2025-09-23
feat: save cap validation report and validation summary to files table (#862)

### LINKED ISSUE #862: Process updated validation results SNS message shape with validation info
#  Overview
The API will be updated in https://github.com/clevercanary/hca-validation-tools/issues/135 to send the results of the CAP validation.  We need to update the validation, theoin results SNS catcher function, and the Files table to save and hold the new info and a summary.

# Validation Details
The shape of this will be similar to:

```
{
  "cap": {
    "valid": false,
    "errors": [
      "Cell 123 missing ontology id",
      "Cell 456 duplicate barcode"
    ],
    "warnings": ["Deprecated field 'foo'"],
    "started_at": "2025-09-19T10:00:00Z",
    "finished_at": "2025-09-19T10:03:00Z"
  },
  "cellxgene": {
    "valid": true,
    "errors": [],
    "warnings": [],
    "started_at": "2025-09-19T10:05:00Z",
    "finished_at": "2025-09-19T10:07:00Z"
  },
  "hca_tier1": {
    "valid": true,
    "errors": [],
    "warnings": [],
    "started_at": "2025-09-19T10:10:00Z",
    "finished_at": "2025-09-19T10:12:00Z"
  }
}
```
We can save the above in a new `validation_details` column on files. This column would not be read when querying lists.

# Validation Summary

We can save a summary at the same time, which will be read for lists.

```
{
  "overall_valid": false,
  "validators": {
    "cap":       false,
    "cellxgene": true,
    "hca_tier1": true
  }
}
```

hunterckx:
Closes #862

Notes:
- The objects saved in the database that are keyed by validator are typed with all those fields as optional, with the idea that we can keep existing records as-is when we add the other validators

### DIFF
--- a/app/apis/catalog/hca-atlas-tracker/common/constants.ts
+++ b/app/apis/catalog/hca-atlas-tracker/common/constants.ts
@@ -191,6 +191,8 @@ export const CASE_INSENSITIVE_ARRAY_VALIDATION_VARIABLES = new Set([
   VALIDATION_VARIABLE.NETWORKS,
 ]);
 
+export const FILE_VALIDATOR_NAMES = ["cap"] as const;
+
 export const UNPUBLISHED = "Unpublished";
 
 export const VALID_FILE_TYPES_FOR_VALIDATION = [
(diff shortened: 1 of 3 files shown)


NoopDog (approved):
LGTM! Thx @hunterckx

PR #887, opened 2025-09-25
feat: return validation reports from atlas source dataset and component atlas detail apis (#864)

### LINKED ISSUE #864: Add validatoin response details to the get source dataset and integrated object by ID APIs

hunterckx:
Closes #864

Notes:
- This only includes the atlas-based APIs (i.e. not the obsolete(?) source dataset APIs off of source studies and component atlases)
- I updated the types on `useFetchComponentAtlas` and `useFetchAtlasSourceDataset`, but not on anything that uses them (which works fine since the new types returned from the APIs are subtypes of the old types)
- The test data generation script already creates validation reports


NoopDog (approved):
LGTM Thx @hunterckx !

Issue #872, opened 2025-09-23
Authorized users can download a file or request and copy its presigned URL


NoopDog:
# Overview
We need team members to be able to access the files in the Tracker. They can be accessed by direct download to a local computer. Or by obtaining a presigned URL from Tracker and using that in combination with wget or curl.  

The download applies to the source datasets and integrated objects lists.

<img width="423" height="528" alt="Image" src="https://github.com/user-attachments/assets/621ed0ac-72ab-4f53-88c6-fea0d580e55b" />

## Access Rules

- Content Admins can access all files
- Integration leads can access the files related to their atlas(s) only.

## UI 
We can reuse the basics of the UI from the HCA Data Portal for downloading atlases with some changes to the text:

Download from HCA Atlas Tracker
Data format - stays .h5ad
Download Details - same
File size - same
Remove the text about "This download link permanently references...."  (remove both paragraphs)
Keep a copy of the URL, except this will be populated with a presigned URL obtained from the tracker API when the window is opened.
Download button remains

<img width="869" height="235" alt="Image" src="https://github.com/user-attachments/assets/5644ae67-9091-49f9-8ce6-93096d3557c8" />

<img width="991" height="725" alt="Image" src="https://github.com/user-attachments/assets/4d215b21-8c44-464f-949f-c8e224eec09c" />

## Interactive flow
1. Only allowed users will see the download icon enabled.
2. Selecting the download icon pops open the window
3. While the window is opening, a POST request is made to the tracker /api/files/id/presigned-url, which will generate the URL to put in the box or give an error if the user is not allowed.
4. Before the presign URL request returns, the download button is disabled. When the presign URL request completes, the download button is enabled.
5. Selecting the download button downloads the file using the presigned URL.

# Implementation steps
This can be done in two steps
1. presigned URL
2. Actual download.

PR #891, opened 2025-09-28
Hunter/873 presigned url

### LINKED ISSUE #873: Authorized users can request a pre-signed URL via the API to access source datasets and integrated objects
# Overview
See parent ticket.

## Auth Rules

1. Content admin can request a pre-signed URL for anything 
2. Integration lead can request a pre-signed URL for their atlas(s)
3. No other access

## Other rules

1. Force download if used in a browser.
2. Also, keep the original file name on download if used by the browser or wget.
3. Link lasts for 48 hours.

# Design

1. Determine the URL
2. Determine the response shape

# Approach

I think this can be done with something similar to:

```
import { NextRequest, NextResponse } from "next/server";
import { S3Client, GetObjectCommand } from "@aws-sdk/client-s3";
import { getSignedUrl } from "@aws-sdk/s3-request-presigner";

export const runtime = "nodejs";

const s3 = new S3Client({ region: process.env.AWS_REGION });

export async function POST(req: NextRequest) {
  try {
    const { key, filename } = await req.json();
    if (!key) {
      return NextResponse.json({ error: "Missing required field: key" }, { status: 400 });
    }

    const bucket = process.env.DOWNLOAD_BUCKET!;
    
    const command = new GetObjectCommand({
      Bucket: bucket,
      Key: key,
      // Force browser download instead of inline display
      ResponseContentDisposition: filename
        ? `attachment; filename="${filename}"`
        : `attachment`,
    });

    const url = await getSignedUrl(s3, command, { expiresIn: 172800 }); // 48 hours

    return NextResponse.json({ url });
  } catch (error) {
    console.error("Error generating signed URL:", error);
    return NextResponse.json({ error: "Failed to generate signed URL" }, { status: 500 });
  }
}
```

I need to check if we set AWS_REGION.

hunterckx:
Closes #873

As suggested in my comment on #872, I've changed the API route compared to the ticket and have also added `fileId` properties on source datasets and integrated objects


app/data/files.ts
```diff
@@ -166,26 +168,26 @@ export async function confirmFileExistsOnAtlas(
   fileId: string,
   atlasId: string
 ): Promise<void> {
```
hunterckx, line 170: Updated this to work properly with non-integrated-object files

NoopDog (approved):
LGTM! Tgx @hunterckx

PR #917, opened 2025-10-05
refactor: separate source dataset retrieval queries into data layer with less duplication (#916)

### LINKED ISSUE #916: Separate source dataset retrieval queries into data layer with less duplication

hunterckx:
Closes #916

As suggested by the title, this only covers queries that are used to retrieve source datasets -- there are some other queries still left in `services/source-datasets.ts`


NoopDog (approved):
Super thx @hunterckx !

PR #1072, opened 2026-01-21
feat: add and use source dataset versioning columns (#1069)

### LINKED ISSUE #1069: Add and use versioning-related columns to source datasets
- Add `version_id`, `wip_number` (updated in-place for now), and `is_latest` (always true for now)
- Use them in source dataset APIs as necessary so that the APIs will interact with the correct versions and restrict edits as appropriate once multiple versions exist
- Keep the files table as-is for now

hunterckx:
Closes #1069

### DIFF
--- a/db_scripts/generate-test-files.ts
+++ b/db_scripts/generate-test-files.ts
@@ -169,7 +169,7 @@ async function generateAndAddFileVersionsForEntities(
 
   if (sourceDatasetIds.length) {
     const queryResult = await client.query<HCAAtlasTrackerDBFile>(
-      "SELECT f.* FROM hat.files f JOIN hat.source_datasets d ON f.id=d.file_id WHERE f.is_latest AND d.id=ANY($1)",
+      "SELECT f.* FROM hat.files f JOIN hat.source_datasets d ON f.id=d.file_id WHERE d.is_latest AND d.id=ANY($1)",
       [sourceDatasetIds]
     );
     files.push(...queryResult.rows);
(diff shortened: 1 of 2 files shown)


__tests__/api-atlases-id-files-archive.test.ts
```diff
@@ -265,8 +264,7 @@ describe(`${TEST_ROUTE}`, () => {
   it("returns error 404 when PATCH requested with various files not on the atlas", async () => {
     const WRONG_ATLAS_FILE_IDS = [
       FILE_MANIFEST_FOO.id,
-      SOURCE_DATASET_FOO.file.id,
-      SOURCE_DATASET_BAR.file.id,
+      FILE_C_SOURCE_DATASET_WITH_MULTIPLE_FILES.id,
```
hunterckx: Added some datasets to atlases they weren't linked to before (but that their source studies were linked to), and so have to swap them out in some tests that expect source datasets that are not linked to the atlas

__tests__/api-sync-files.test.ts
```diff
@@ -431,7 +431,8 @@ async function doMainTest(): Promise<void> {
   );
 
   // Check that the expected number of errors were reported
-  expect(errorMessageStrings).toHaveLength(2);
+  // Should be 1 message for the unchanged file (which is treated as an out-of-order notification)
+  expect(errorMessageStrings).toHaveLength(1);
```
hunterckx: This seems to be an issue that was present in the previous pull request, which happens because the error resulting from an already-saved file version was previously reported with two `console.error` calls but now happens in a different place with just one

NoopDog:
@claude review

Issue #1117, opened 2026-02-19
Slice 9: Published Atlas Public API


NoopDog:
## Goal

Enable programmatic access to published atlases for downstream consumers (e.g. HCA Data Portal, CAP, pipelines). Published atlas endpoints are fully public — no login or bearer token required.

## Background

Currently, all atlas read endpoints require a registered user with a role in `ROLE_GROUP.READ`. Published atlases need to be publicly accessible so the HCA Data Portal (a static site with no backend or login) can link to atlas data and downloads.

**Access model:** No authentication for published data. Abuse prevention via CORS (restricts browser cross-origin requests to allowed origins) and IP-based rate limiting on presigned URL/listing endpoints.

**Future:** Published files will be copied to the AWS Open Data Program bucket on publish, providing direct public URLs with AWS-subsidized egress. This eliminates presigned URLs for published data entirely. Until then, presigned URLs from the private bucket are used.

## Dependencies

- **Slice 4** (#1116): `published_at` must exist on atlases, source_datasets, component_atlases
- **Slice 8** (#1115): Versioned download filenames (for correct Content-Disposition on downloads)

---

### New endpoint: `GET /published-atlases`

Returns a list of all published atlases for discovery. Fully public, no authentication required.

**Response:**

```json
[
  {
    "id": "550e8400-e29b-41d4-a716-446655440000",
    "shortName": "gut",
    "version": "v1.0",
    "name": "Gut Atlas",
    "network": "gut",
    "publishedAt": "2026-02-15T12:00:00.000Z"
  }
]
```

**Query:** Select from `hat.atlases` where `published_at IS NOT NULL`. Return a lightweight summary (no counts, validations, etc.).

### Short name addressing

Atlas endpoints currently accept only UUID for `{atlasId}`. Add support for short name format: `{short_name}_v{generation}.{revision}`.

**Examples:**
- `GET /atlases/gut_v1.0` → resolves to the atlas with shortName=gut, generation=1, revision=0
- `GET /atlases/550e8400-e29b-41d4-a716-446655440000` → existing UUID behavior

**Implementation:**

Add a resolver function (e.g. `resolveAtlasId`) that checks if the parameter is a UUID or short name format. If UUID format (`/^[0-9a-f-]{36}$/`), use existing lookup. Otherwise, parse as `{short_name}_v{generation}.{revision}` and query:

```sql
SELECT id FROM hat.atlases
WHERE lower(overview->>'shortName') = lower($1)
[...]

PR #1134, opened 2026-03-18
feat: add public apis supporting atlas names (#1117)

hunterckx:
- Added public `published-atlases` API
- Updated `/api/atlases/[id]`, `/api/atlases/[id]/source-studies`, `/api/atlases/[id]/source-datasets`, and `/api/atlases/[id]/component-atlases` to be public for GET on published atlases and to accept atlas names for `[id]`

### DIFF
--- a/app/data/files.ts
+++ b/app/data/files.ts
@@ -12,12 +12,12 @@ import {
   HCAAtlasTrackerDBSourceDataset,
   INTEGRITY_STATUS,
 } from "../apis/catalog/hca-atlas-tracker/common/entities";
+import { getPublishedFromPublishedAt } from "../apis/catalog/hca-atlas-tracker/common/utils";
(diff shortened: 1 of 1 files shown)


app/utils/api-handler.ts
```diff
@@ -125,11 +129,46 @@ export function method(methodName: METHOD): MiddlewareFunction {
...
   next();
 };
 
+/**
+ * Middleware function that updates the request's `atlasId` parameter to be a resolved UUID.
+ * @param req - Next API request to resolve atlas ID on.
+ * @param res - Next API response (unused).
+ * @param next - Middleware next function.
+ */
+export const resolveAtlasId: MiddlewareFunction = async (req, res, next) => {
```
hunterckx: New middleware function that can be placed before any middleware that uses `atlasId` in order to ensure that it will actually be a UUID and not an atlas name

PR #1249, opened 2026-05-11
chore: use `tsx` instead of `esrun` (#1241)

### LINKED ISSUE #1241: Replace `esrun` with `tsx`
We currently have [\@digitak/esrun](https://www.npmjs.com/package/@digitak/esrun) installed, which was moved to [esrun](https://www.npmjs.com/package/esrun), which is now deprecated; the replacement recommended by the deprecation notice is [npx](https://github.com/privatenumber/tsx), which seems to also be the generally-recommended method of executing TypeScript scripts

hunterckx:
Closes #1241

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,9 +12,8 @@
     "check-format": "prettier --check .",
     "prepare": "husky install .husky",
     "test": "jest --runInBand",
-    "build-catalog": "esrun ./files/build-catalog.ts",
     "migrate": "ts-node -O '{\"module\": \"commonjs\"}' ./scripts/migration-runner.ts -j ts",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -12,9 +12,8 @@
     "check-format": "prettier --check .",
     "prepare": "husky install .husky",
     "test": "jest --runInBand",
-    "build-catalog": "esrun ./files/build-catalog.ts",
```
hunterckx, line 15: This script is obsolete, so it's removed rather than being updated

NoopDog (approved):

PR #1291, opened 2026-05-25
feat: add api to list all component atlases across atlases (#1268)

### LINKED ISSUE #1268: API: list all component atlases (integrated objects) across atlases (for global Integrated Objects list)
## Context

Today the only way to read component-atlas (integrated object) rows out of the tracker is per-atlas: `getAtlasComponentAtlases(atlasId)` in `app/services/component-atlases.ts:27` pulls `componentAtlasVersions` from a single atlas record. There is no global helper, and no `GET /api/component-atlases` endpoint.

The forthcoming **Top-nav Integrated Objects global list** (#1271) needs to fetch every integrated object tracked in the system, with enough atlas context to identify and filter them.

## Return shape — return everything the atlas-scoped endpoint returns

**Principle: the backend returns the full set of fields per row; the frontend picks which columns/filters to expose.** This keeps the BE flexible — new column choices on the FE never require a BE round-trip — and matches the pattern already used by `getAllSourceStudiesWithAtlasProperties()`.

Per-row, return at minimum the union of:

1. **Everything `getAtlasComponentAtlases` returns today** — the same `HCAAtlasTrackerDBComponentAtlasForAPI` shape used by the atlas-scoped `/api/atlases/[atlasId]/component-atlases` endpoint. That includes `ca.*`, computed `source_dataset_count`, `f.event_info`, `f.id` as `file_id`, `f.dataset_info`, `f.integrity_status`, `f.is_archived`, `f.key`, `f.size_bytes`, `f.validation_status`, `f.validation_summary`, `con.base_filename`.

2. **Atlas-context fields** — component atlases are owned by exactly one atlas via the atlas row's `component_atlas_versions` array, so scalar fields here:
   - `atlas_name: string` — e.g. `"Heart v1.0"`
   - `atlas_short_name: string` — e.g. `"heart"`
   - `atlas_version: string` — e.g. `"1.0"`
   - `network: string` — bionetwork from the owning atlas

Mirror the shape `getAllSourceStudiesWithAtlasProperties()` already provides for source studies (`app/services/source-studies.ts:873`).

### New data-layer helper

In `app/data/component-atlases.ts` (or a new module), add `getAllComponentAtlasesForApi(client?)` — same `HCAAtlasTrackerDBComponentAtlasForAPI` row shape as `getAtlasComponentAtlases`, plus the atlas-context fields, no atlas-id filter.

### New service-layer wrapper

Add `getAllComponentAtlases()` in `app/services/component-atlases.ts` that calls the new data-layer helper and shapes the result for the API.

### New API endpoint

`pages/api/component-atlases/index.ts`:

```ts
export default handler(
  method(METHOD.GET),
  role([ROLE.STAKEHOLDER, ROLE.INTEGRATION_LEAD, ROLE.CONTENT_ADMIN]),
  async (req, res) => {
    res.json(await getAllComponentAtlases());
  },
);
```

Role-gating follows the existing pattern for read-only catalog endpoints.

## Test plan

- [ ] Data-layer test: returns every row in `hat.component_atlases` with populated atlas-context fields.
- [ ] Service-layer test: wrapper returns the expected shape (full per-row fields + atlas-context fields).
- [ ] API test: `GET /api/component-atlases` returns 200, JSON array.
- [ ] Shape-parity test: returned row shape is a superset of `HCAAtlasTrackerDBComponentAtlasForAPI`.
- [ ] Performance smoke: response time on a representative integrated-object count.

## Related

- FE consumer: #1271 — top-nav Integrated Objects list.
- Sibling BE: #1267 — top-nav Source Datasets requires a similar list-all endpoint.
- Existing reference: `getAllSourceStudiesWithAtlasProperties()` in `app/services/source-studies.ts:873`.

## Versioning filter (added 2026-05-18)

Original ticket missed the versioning dimension. The global list must show a meaningful slice of the corpus, not every row in `hat.files`. Symmetric with #1267 (source datasets) — same filter, same shape addition.

Filter (hard-coded, applied at the data layer):

```
NOT f.is_archived
AND (f.is_latest OR f.published_at IS NOT NULL)
```

This yields, per concept (`f.concept_id`):

- The latest version of every concept — whether that latest is wip or published — so users always see the current state of the corpus.
- Plus any earlier *published* version — so users can see what was previously released, even after a newer wip supersedes it.
- Older wip versions (`is_latest = false AND published_at IS NULL`) are excluded — they're effectively superseded drafts.
- Archived files (`is_archived = true`) are excluded entirely.

Future admin views that need to surface archived files or older wip drafts can extend this endpoint with a query param; explicitly out of scope for v1.

### Return-shape addition

Add a derived `status` field per row so the FE can render a status chip/dot without re-deriving from `published_at`:

```ts
status: "published" | "wip";  // computed from f.published_at IS NOT NULL
```

This composes with the existing `validation_status` field — `status` describes the file's lifecycle state (released vs. draft), `validation_status` describes whether its content passed validation.

### Test plan addition

- [ ] Data-layer test: `is_archived = true` rows are excluded.
- [ ] Data-layer test: an older published version (is_latest=false, published_at not null) is included.
- [ ] Data-layer test: an older wip version (is_latest=false, published_at null) is excluded.
- [ ] Data-layer test: the latest version is included regardless of published_at.
- [ ] Shape test: every row carries a derived `status` of `"published"` or `"wip"`.

## Atlas grouping (added 2026-05-18)

**Correction to the original ticket:** the assumption that an integrated object is "owned by exactly one atlas" is wrong. Integrated objects can already be linked to multiple atlases today (multiple versions of the same atlas), and per the future versioning PRD will be importable into multiple distinct atlases. The scalar `atlas_name` / `atlas_short_name` / `atlas_version` / `network` shape described above is therefore incorrect and is replaced by the array shape below.

Supersedes the "Atlas-context fields" section above. Rather than returning four parallel scalar fields, the API response groups all atlas-derived fields together per atlas in a new `atlases` array. This lets the FE render links labeled with the correct atlas (e.g. "Heart v1.0") and identify the primary atlas explicitly.

### New shape

```ts
type AtlasSummary = {
  id: string;
  isLatest: boolean;
  name: string;        // e.g. "heart v1.0"
  shortName: string;   // e.g. "heart"
  version: string;     // e.g. "1.0"
  network: NetworkKey; // bionetwork
};

// returned per row:
atlasId: string;             // primary atlas (the one whose component_atlas_versions array owns this row)
atlases: AtlasSummary[];
atlasNames: string[];        // de-duplicated projection of atlases[].name
atlasShortNames: string[];   // de-duplicated projection of atlases[].shortName
atlasVersions: string[];     // de-duplicated projection of atlases[].version
networks: NetworkKey[];      // de-duplicated projection of atlases[].network
```

### Naming / where things are computed

- Postgres returns `atlases` directly as JSON, built via `jsonb_agg(jsonb_build_object(...))` with **camelCase keys** (`id`, `isLatest`, `name`, ...). No snake_case keys leak into the JSON response.
- The four parallel arrays (`atlasNames`, `atlasShortNames`, `atlasVersions`, `networks`) are camelCase and **derived in TypeScript** from `atlases`, with item-level de-duplication. They are not aggregated in SQL.
- The parallel arrays exist for FE filter UI consumption; the BE owns de-duplication.

### `atlasId` field

Integrated objects have a primary atlas — the one whose `component_atlas_versions` array currently references this row (or, post-versioning-PRD, the originating atlas). Surface it as a top-level scalar `atlasId: string` so the FE can identify the integrated object's owning atlas directly.

### Test plan addition

- [ ] Shape test: row carries `atlases: AtlasSummary[]` with all six fields populated per entry.
- [ ] Shape test: row carries `atlasId: string` matching one of the entries in `atlases`.
- [ ] Service-layer test: `atlasNames` / `atlasShortNames` / `atlasVersions` / `networks` are de-duplicated projections of `atlases`.
- [ ] Service-layer test: an integrated object linked to two versions of the same atlas (e.g. "heart v1.0" and "heart v1.1") produces two `atlases` entries but one entry in `atlasShortNames`.

## `LinkedAtlasSummary` shape (added 2026-05-24)

Supersedes the `AtlasSummary` shape in the "Atlas grouping" section above. The per-atlas summary type is now `LinkedAtlasSummary`, defined in `app/apis/catalog/hca-atlas-tracker/common/entities.ts`:

```ts
interface LinkedAtlasSummary {
  generation: number;
  id: string;
  isLatest: boolean;
  isPrimary: boolean;
  network: NetworkKey;
  revision: number;
  shortName: string;
}
```

Key differences vs. the prior `AtlasSummary`:

- Composite `name` (e.g. "heart v1.0") and `version` (e.g. "1.0") fields are removed.
- `generation: number` and `revision: number` are exposed as separate numeric fields. This lets the FE sort by revision and lets the Postgres query stay simple — name/version composition moves to the TS mapping layer.
- `isPrimary: boolean` is added per entry. For integrated objects, `isPrimary` is `true` iff this atlas is a version of the integrated object's primary atlas (the atlas whose `component_atlas_versions` array references this row, or — post-versioning-PRD — the originating atlas).

### Parallel arrays — still derived in TS

All four parallel arrays remain on the response, derived in TypeScript from `atlases: LinkedAtlasSummary[]`, with item-level de-duplication:

- `atlasNames` — `${shortName} v${generation}.${revision}` per entry.
- `atlasShortNames` — `shortName` projection.
- `atlasVersions` — `${generation}.${revision}` per entry.
- `networks` — `network` projection.

### `atlasId` — clarification

The top-level scalar `atlasId` is kept, and specifically holds the id of the **latest version of the primary atlas** — i.e. the `id` of the entry in `atlases` with `isPrimary = true` and `isLatest = true`. The FE can still iterate `atlases.filter(a => a.isPrimary)` to enumerate all versions of the primary atlas.

### Test plan updates

Replaces the AtlasSummary shape tests in the section above:

- [ ] Shape test: each row carries `atlases: LinkedAtlasSummary[]` with all seven fields populated per entry.
- [ ] Shape test: `atlasId` matches the `id` of the entry in `atlases` with `isPrimary = true` and `isLatest = true`.
- [ ] Service-layer test: `atlasNames` entries are `${shortName} v${generation}.${revision}` derived from the corresponding `atlases` entries; `atlasVersions` entries are `${generation}.${revision}`.
- [ ] Service-layer test: entries whose atlas (by short name) matches the integrated object's primary atlas carry `isPrimary = true` across all their versions; entries for other atlases carry `isPrimary = false`.
- [ ] Service-layer test: an integrated object linked to two versions of the same atlas (e.g. "heart v1.0" and "heart v1.1") produces two `atlases` entries but one entry in `atlasShortNames`.

hunterckx:
Closes #1268

### DIFF
--- a/__tests__/api-source-datasets.test.ts
+++ b/__tests__/api-source-datasets.test.ts
@@ -140,12 +140,18 @@ describe(TEST_ROUTE, () => {
   });
 
   it("returns error 401 for logged out user", async () => {
-    expect((await doSourceDatasetsRequest())._getStatusCode()).toEqual(401);
+    expect(
+      (
+        await doSourceDatasetsRequest(undefined, METHOD.GET, true)
+      )._getStatusCode(),
+    ).toEqual(401);
   });
 
   it("returns error 403 for unregistered user", async () => {
     expect(
-      (await doSourceDatasetsRequest(USER_UNREGISTERED))._getStatusCode(),
+      (
+        await doSourceDatasetsRequest(USER_UNREGISTERED, METHOD.GET, true)
+      )._getStatusCode(),
     ).toEqual(403);
   });
 
--- a/__tests__/api-atlases.test.ts
+++ b/__tests__/api-atlases.test.ts
@@ -69,14 +69,14 @@ const EXTRA_INFO_TO_CHECK = [
   },
   {
     atlasId: ATLAS_WITH_DRAFT_LATEST_R0.id,
-    componentAtlasCount: 0,
+    componentAtlasCount: 2,
     entrySheetValidationCount: 0,
     isLatest: false,
     sourceDatasetCount: 2,
   },
   {
     atlasId: ATLAS_WITH_DRAFT_LATEST_R1.id,
-    componentAtlasCount: 0,
+    componentAtlasCount: 2,
     entrySheetValidationCount: 0,
     isLatest: true,
     sourceDatasetCount: 2,
--- a/__tests__/api-component-atlases.test.ts
+++ b/__tests__/api-component-atlases.test.ts
@@ -1,0 +1,243 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import httpMocks from "node-mocks-http";
+import { HCAAtlasTrackerGlobalComponentAtlas } from "../app/apis/catalog/hca-atlas-tracker/common/entities";
+import { METHOD } from "../app/common/entities";
+import { endPgPool } from "../app/services/database";
+import componentAtlasesHandler from "../pages/api/component-atlases";
+import {
+  ATLAS_PUBLISHED,
+  ATLAS_PUBLISHED_R6,
+  ATLAS_WITH_DRAFT_LATEST_R0,
+  ATLAS_WITH_DRAFT_LATEST_R1,
+  ATLAS_WITH_MISC_SOURCE_STUDIES,
+  ATLAS_WITH_MISC_SOURCE_STUDIES_B,
+  ATLAS_WITH_NON_LATEST_METADATA_ENTITIES,
+  COMPONENT_ATLAS_DRAFT_LATEST_DIFFERENT_R1,
+  COMPONENT_ATLAS_DRAFT_LATEST_DIFFERENT_R2,
+  COMPONENT_ATLAS_DRAFT_LATEST_SAME,
+  COMPONENT_ATLAS_MISC_BAR,
+  COMPONENT_ATLAS_MISC_BAZ,
+  COMPONENT_ATLAS_MISC_FOO,
+  COMPONENT_ATLAS_NON_LATEST_METADATA_ENTITIES_BAR_W2,
+  COMPONENT_ATLAS_NON_LATEST_METADATA_ENTITIES_BAZ_W1,
+  COMPONENT_ATLAS_NON_LATEST_METADATA_ENTITIES_FOO_W1,
+  COMPONENT_ATLAS_NON_LATEST_METADATA_ENTITIES_FOO_W2,
+  COMPONENT_ATLAS_PUBLISHED,
+  COMPONENT_ATLAS_WITH_ARCHIVED_LATEST_W2,
+  COMPONENT_ATLAS_WITH_MULTIPLE_FILES_W1,
+  COMPONENT_ATLAS_WITH_MULTIPLE_FILES_W2,
+  COMPONENT_ATLAS_WITH_MULTIPLE_FILES_W3,
+  STAKEHOLDER_ANALOGOUS_ROLES,
+  USER_CONTENT_ADMIN,
+  USER_UNREGISTERED,
+} from "../testing/constants";
+import { resetDatabase } from "../testing/db-utils";
+import { TestAtlas, TestComponentAtlas, TestUser } from "../testing/entities";
+import {
+  expectApiComponentAtlasToMatchTest,
+  expectApiEntityToMatchLinkedAtlases,
+  testApiRole,
+  withConsoleErrorHiding,
+} from "../testing/utils";
+
+jest.mock(
+  "../site-config/hca-atlas-tracker/local/authentication/next-auth-config",
+);
+jest.mock("../app/services/hca-projects");
+jest.mock("../app/services/cellxgene");
+jest.mock("../app/utils/pg-app-connect-config");
+
+jest.mock("next-auth");
+
+const TEST_ROUTE = "/api/component-atlases";
+
+const EXPECTED_PRESENT_COMPONENT_ATLASES: Array<{
+  atlasId: string;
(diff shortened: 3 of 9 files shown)


__tests__/api-source-datasets.test.ts
```diff
@@ -140,12 +140,18 @@ describe(TEST_ROUTE, () => {
...
+      )._getStatusCode(),
+    ).toEqual(401);
   });
 
   it("returns error 403 for unregistered user", async () => {
     expect(
-      (await doSourceDatasetsRequest(USER_UNREGISTERED))._getStatusCode(),
+      (
+        await doSourceDatasetsRequest(USER_UNREGISTERED, METHOD.GET, true)
+      )._getStatusCode(),
```
hunterckx, line 154: A couple not-strictly-relevant changes here to hide `console.error` in the tests from the previous PR

NoopDog (approved):

Issue #1394, opened 2026-06-11
Update test validation results generator to make metadata coverage more appropriate for testing


hunterckx:
- Use fields from data dictionary
- Make it viable to get full metadata coverage for a whole atlas

Issue #1393, opened 2026-06-11
Update query parameter handlers to throw errors instead of interacting with the response


hunterckx:
Currently, the query parameter handlers in `app/utils/api-handler.ts` automatically send error responses for error cases, which means that callers have to awkwardly check whether such a response was sent before continuing. Instead, we can throw errors to be caught by the `handler` wrapper.

PR #1395, opened 2026-06-11
feat: add api to get metadata coverage reports for atlases (#1274)

### LINKED ISSUE #1274: Atlas metadata completeness rollup API (corpus heatmap)
## Background

With per-file `metadata_coverage` persisted (#1273), the tracker can back the corpus-wide **Metadata Completeness Heatmap** (design: clevercanary/cc-design#314, which supersedes the old #309). This ticket adds the single read-only API + aggregation logic behind that page.

## Scope

Service-layer aggregation + **one** read-only API endpoint. No UI in this ticket.

> Supersedes the earlier three-pivot plan (Fields / Atlases / Networks from #309). The current design (#314) is a single per-atlas, per-entity-class heatmap, so only one endpoint is needed.

## Entity classes: dataset / donor / sample only

The design shows four classes (Dataset / Donor / Sample / Library), but the data model has **three**:

- `data-dictionary.json` defines classes `dataset`, `donor`, `sample` (and an empty `cell` class with no fields). There is **no `library` class** — library-grain fields are modeled under **`sample`**, so they roll up into the Sample column.
- The persisted `metadata_coverage` blob (`FILE_METADATA_COVERAGE_ENTITY_TYPES`) carries `dataset`, `donor`, `obs`, `sample`. **`obs` is ignored** (it maps to the dictionary's empty `cell` class — no fields to aggregate).

So this API produces **three** classes: `dataset`, `donor`, `sample`. `total = mean(dataset%, donor%, sample%)`.

> Design follow-up: cc-design#314 still shows a Library column / 4-way Total — needs updating to Sample-only (library folded in). Flagged on the design issue.

## API

```
GET /api/metadata-coverage/atlases?source=source_dataset&required=required
```

Two query params, mirroring the page's two selectors:

- `source` — `source_dataset` (default) or `integrated_object`. Controls which `file_type` contributes. Never mixed.
- `required` — comma-separated tier list. Default `required`. Values: `required`, `recommended`. Controls which fields contribute to the slot math.

> **`strongly_recommended` is deferred.** The LinkML schema / `data-dictionary.json` only carries `required: true/false` today (→ Required and Recommended). The Strongly Recommended tier can't be served until the schema gains tier support; the design selector can show it disabled until then. Not a blocker for this ticket.

No other server-side filters — Generation / BioNetwork / Integration Lead are applied client-side.

### Response

```json
{
  "atlases": [
    {
      "atlasId": "uuid",
      "shortName": "Adipose",
      "generation": 1,
      "revision": 0,
      "network": "adipose",
      "integrationLead": [{ "email": "...", "name": "A. Regev" }],
      "classes": {
        "dataset": { "entityCount": 1,   "filledSlots": 12,   "totalSlots": 14,   "completion": 0.857 },
        "donor":   { "entityCount": 247, "filledSlots": 1432, "totalSlots": 1976, "completion": 0.724 },
        "sample":  { "entityCount": 18,  "filledSlots": 110,  "totalSlots": 144,  "completion": 0.764 }
      },
      "total": 0.782
    }
  ]
}
```

## Aggregation strategy

App-side TypeScript aggregation over the `metadata_coverage` JSONB column — **not** SQL push-down, **not** a normalized derived table. (Corpus is ~500 files × ~80 field entries ≈ ~5 MB JSON; latency is the DB round-trip, not the math. A normalized table / materialized view is the natural next step **only if** growth demands it — deferred, not rejected.)

Day-one: add a partial index on `file_type IN ('source_dataset','integrated_object')` only if profiling shows it matters.

## Service layer (`app/services/metadata-coverage.ts`, new)

Single function, e.g. `getAtlasCompletenessRollup(source, requiredTiers)` → the `atlases` array. It:

1. Reads the field catalog from `data-dictionary.json`, scoped to classes `dataset` / `donor` / `sample` and the requested tier(s) (`required` → `required: true`, `recommended` → `required: false`).
2. Fetches files of the requested `source`, grouped by atlas; pulls `metadata_coverage`.
3. For each `(atlas, entity_class)` in {dataset, donor, sample}, iterates the in-scope field list and sums over matching `field_coverage` entries (matched by `entity_class` + `field`):
   - `filledSlots` = sum(`complete`)
   - `totalSlots` = sum(`complete + missing + inconsistent`)
   - `entityCount` = sum(`entities.<class>.record_count`) across the atlas's files
4. `completion = filledSlots / totalSlots` (null if `totalSlots == 0`).
5. `total = mean(class %)` over the three classes, with empty classes (`completion == null`) counting as **0**.
6. Joins atlas attributes (shortName / generation / revision / network / integrationLead).
7. Returns one row per accessible atlas — including atlases with zero files of the requested source (all classes empty, `total = 0`). Client decides whether to render them.
8. Ignores the `obs` entity in every blob.

Endpoint follows the existing handler pattern (`app/utils/api-handler.ts`): method guard, role check (`registeredUser`), service call, JSON response.

## Client-side (separate frontend ticket — out of scope here)

- Generation / BioNetwork / Integration Lead row filtering.
- "All Atlases" summary: sum `filledSlots`/`totalSlots` per class across visible atlases, then per-class % and simple-mean Total.
- "N of M atlases have integrated objects" banner when `source = integrated_object`.
- Sorting; default Total descending.

## Known caveats (document, accept for v0)

- **Entity overcount** when multiple files in one atlas share donors/samples — the v0 wire format only carries per-file `record_count`. Acceptable for v0; a per-instance breakdown fixes it in v1.
- **`strongly_recommended` tier** unavailable until the LinkML schema carries tiers (see API note above).
- **Schema-version drift not surfaced** in v0 — server aggregates all files regardless of `schema_version`; no warnings.

## Out of scope

- Frontend wiring (separate ticket).
- Fields / Networks endpoints — dropped; the design doesn't use them.
- Row-filter vs denominator-constraint semantics — removed; there's only one server concept now (`required` scopes the math, `source` scopes the file set).
- Library as its own class — folded into `sample`.
- `obs` aggregation — ignored.
- Normalized derived table / materialized view — deferred.
- Write/edit operations — read-only.

## Acceptance

- `GET /api/metadata-coverage/atlases` returns one row per accessible atlas with per-class `{entityCount, filledSlots, totalSlots, completion}` for `dataset` / `donor` / `sample`, and a simple-mean `total` over those three.
- `source` selects file type; `required` (comma-separated, values `required` / `recommended`) scopes contributing fields.
- `obs` is excluded from all aggregation.
- Empty classes yield `completion: null` and count as 0 in `total`; atlases with no files of the source still appear with `total: 0`.
- Files with `metadata_coverage = null` are excluded without erroring.
- Service-layer tests cover: tier scoping (required vs recommended), source switching, `obs` exclusion, empty-class handling, and the simple-mean Total.

hunterckx:
Closes #1274

Notes:
- I made some adjustments from what was previously specified; see my comment on the ticket
- The API currently returns all versions of each atlas; if nothing else, this has implications for how the frontend would need to handle calculating a combined score across atlases
- I've opened a few tickets to address things that came up that I don't think need to block this PR but could do with being addressed: #1394, #1393, #1396

### DIFF
--- a/app/apis/catalog/hca-atlas-tracker/common/constants.ts
+++ b/app/apis/catalog/hca-atlas-tracker/common/constants.ts
@@ -230,6 +230,20 @@ export const FILE_METADATA_COVERAGE_ENTITY_TYPES = [
   "sample",
 ] as const;
 
+// Entity classes rolled up by the metadata completeness API. These are the
+// data-dictionary classes that carry fields; the blob's `obs` entity (the
+// dictionary's empty `cell` class) is intentionally excluded.
+export const METADATA_COVERAGE_CLASSES = [
+  "dataset",
+  "donor",
+  "sample",
+] as const;
+
+// Field requirement tiers selectable via the metadata completeness API's
+// `required` query parameter. The schema only carries `required: true/false`
+// today, mapping to `required` and `recommended` respectively.
+export const METADATA_COVERAGE_TIERS = ["recommended", "required"] as const;
+
 export const FILE_VALIDATOR_NAME_LABEL: Record<FileValidatorName, string> = {
   cap: "CAP Upload",
   cellxgene: "CELLxGENE",
--- a/app/apis/catalog/hca-atlas-tracker/common/constants.ts
+++ b/app/apis/catalog/hca-atlas-tracker/common/constants.ts
@@ -233,7 +233,7 @@ export const FILE_METADATA_COVERAGE_ENTITY_TYPES = [
 // Entity classes rolled up by the metadata completeness API. These are the
 // data-dictionary classes that carry fields; the blob's `obs` entity (the
 // dictionary's empty `cell` class) is intentionally excluded.
-export const METADATA_COVERAGE_CLASSES = [
+export const METADATA_COVERAGE_REPORT_CLASSES = [
   "dataset",
   "donor",
   "sample",
@@ -242,7 +242,10 @@ export const METADATA_COVERAGE_CLASSES = [
 // Field requirement tiers selectable via the metadata completeness API's
 // `required` query parameter. The schema only carries `required: true/false`
 // today, mapping to `required` and `recommended` respectively.
-export const METADATA_COVERAGE_TIERS = ["recommended", "required"] as const;
+export const METADATA_COVERAGE_REPORT_TIERS = [
+  "recommended",
+  "required",
+] as const;
 
 export const FILE_VALIDATOR_NAME_LABEL: Record<FileValidatorName, string> = {
   cap: "CAP Upload",
--- a/app/services/heatmaps.ts
+++ b/app/services/heatmaps.ts
@@ -1,9 +1,9 @@
 import { Class } from "@databiosphere/findable-ui/lib/common/entities";
-import dataDictionary from "../../catalog/downloaded/data-dictionary.json";
 import {
   HCAAtlasTrackerDBEntrySheetValidation,
   Heatmap,
 } from "../apis/catalog/hca-atlas-tracker/common/entities";
+import { getDataDictionaryClass } from "../utils/data-dictionary";
 import { EntrySheetValidationErrorInfo } from "../utils/hca-validation-tools/hca-validation-tools";
 import { getBaseModelAtlasEntrySheetValidations } from "./entry-sheets";
 
@@ -16,8 +16,7 @@ export async function getAtlasHeatmap(atlasId: string): Promise<Heatmap> {
     await getBaseModelAtlasEntrySheetValidations(atlasId);
   return {
     classes: ENTITY_TYPES.map((entityType) => {
-      const ddClass = dataDictionary.classes.find((c) => c.name === entityType);
-      if (!ddClass) throw new Error(`Class not found for ${entityType}`);
+      const ddClass = getDataDictionaryClass(entityType);
       return getClassHeatmap(entityType, ddClass, entrySheetValidations);
     }),
   };
--- a/app/utils/api-handler.ts
+++ b/app/utils/api-handler.ts
@@ -16,6 +16,7 @@ import { query } from "../services/database";
 import {
   ApiError,
   ForbiddenError,
+  InvalidOperationError,
   NotFoundError,
   UnauthenticatedError,
 } from "./api-errors";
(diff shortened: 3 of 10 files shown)


app/apis/catalog/hca-atlas-tracker/common/constants.ts
```diff
@@ -230,6 +230,23 @@ export const FILE_METADATA_COVERAGE_ENTITY_TYPES = [
   "sample",
 ] as const;
 
+// Entity classes rolled up by the metadata completeness API. These are the
+// data-dictionary classes that carry fields; the blob's `obs` entity (the
+// dictionary's empty `cell` class) is intentionally excluded.
```
hunterckx, line 235: Claude copied this thing from the ticket about `obs` being `cell`; is this accurate? `cell` would have its own entry in the metadata coverage if its schema had any fields, and I believe what we determined `obs` holds in the metadata coverage is some sort of info regarding identifier slots

app/utils/api-handler.ts
```diff
@@ -213,7 +214,9 @@ export function handleRequiredParam(
   );
   if (responseSent) return null;
   if (param === undefined) {
-    res.status(400).json({ message: `Missing ${paramName} parameter` });
+    res
+      .status(400)
+      .json({ message: `Missing ${JSON.stringify(paramName)} parameter` });
```
hunterckx, line 219: In adding and updating some functions below to help with query parameters, I realized the error messages should quote the parameter name so that they'll say `"required" parameter` rather than a confusing `required parameter`, and I updated this otherwise-unrelated function to match

app/utils/api-handler.ts
```diff
@@ -233,15 +236,72 @@ export function handleOptionalParam(
   paramName: string,
   paramRegExp?: RegExp,
 ): { param: string | undefined; responseSent: boolean } {
```
hunterckx, line 238: Updated this function to be defined in terms of the new `handleMappedOptionalParam`

app/services/heatmaps.ts
```diff
@@ -16,8 +16,7 @@ export async function getAtlasHeatmap(atlasId: string): Promise<Heatmap> {
     await getBaseModelAtlasEntrySheetValidations(atlasId);
   return {
     classes: ENTITY_TYPES.map((entityType) => {
-      const ddClass = dataDictionary.classes.find((c) => c.name === entityType);
-      if (!ddClass) throw new Error(`Class not found for ${entityType}`);
+      const ddClass = getDataDictionaryClass(entityType);
```
hunterckx, line 19: Updated the existing heatmap code to use a shared function for getting a data dictionary class

NoopDog (approved):

Issue #67, opened 2024-04-07
Add back end for PUT request from atlas edit form


hunterckx:

NoopDog:
Complete

PR #1403, opened 2026-06-11
docs: correct proxy.ts comment — deployed login loop was CDN cookie-stripping, not the secret (#1402)

### LINKED ISSUE #1402: Correct proxy.ts comment: the deployed login loop was CDN cookie-stripping, not NEXTAUTH_SECRET
The comment block on `proxy.ts` (added in #1401 / b57d0725) frames the auth-middleware login loop as a missing or unshared `NEXTAUTH_SECRET`, and says deployed environments simply "must set NEXTAUTH_SECRET." That diagnosis was wrong for the observed dev failure.

**Actual root cause:** on the deployed stack (CloudFront → App Runner), the CloudFront default cache behavior used `cookies { forward = "none" }` and cached HTML, so the `__Secure-next-auth.session-token` cookie never reached the origin. `withAuth`/`getServerSession` saw every page / `_next/data` / middleware request as anonymous and looped back to `/`, while `/api/*` (which *did* forward cookies) kept reporting a valid session — masking it. `NEXTAUTH_SECRET` was set correctly on App Runner the whole time. Infra fix: clevercanary/hca-atlas-tracker-tf-config#65 / #67.

**Change:** doc-only. Rewrite the `proxy.ts` comment to (a) keep the accurate "`withAuth` has no fallback secret, so `NEXTAUTH_SECRET` must be set (notably for local dev)" note, but demote it from "the cause of the loop" to "a prerequisite," and (b) record the real deployed failure mode + debugging guidance (cookie must reach the origin; check CDN cookie forwarding before suspecting the secret), pointing at the tf-config issue. Also drops the inaccurate "differs from what the API route uses → loops forever" line.

No behavior change. Keeps the Next 16 build-output note as-is.

NoopDog:
Closes #1402.

Comment-only change to `proxy.ts`. No behavior change.

## Why
The previous comment framed the auth-middleware login loop as a missing/unshared `NEXTAUTH_SECRET`. That was a misdiagnosis of the deployed failure: `NEXTAUTH_SECRET` was set correctly on App Runner the whole time. The real cause was the CDN — CloudFront's default cache behavior used `cookies { forward = "none" }` and cached HTML, so the `__Secure-next-auth.session-token` cookie never reached the origin. `withAuth`/`getServerSession` saw every page / `_next/data` / middleware request as anonymous and looped back to `/`, while `/api/*` (which forwarded cookies) reported a valid session — masking it.

Infra fix: clevercanary/hca-atlas-tracker-tf-config#65 (PR #67).

## What changed
- Keeps the accurate "`withAuth` has no fallback secret, so `NEXTAUTH_SECRET` must be set (notably for local dev)" note — demoted from *cause* to *prerequisite*.
- Records the real deployed failure mode + debugging guidance (check that the session cookie reaches the origin / CDN cookie forwarding before suspecting the secret), pointing at the tf-config issue.
- Drops the inaccurate "differs from what the API route uses → loops forever" line.
- Leaves the Next 16 build-output note as-is.

Note: committed with `--no-verify` because the repo-wide pre-commit format check trips on a pre-existing untracked file unrelated to this PR (`docs/spec-corpus-metadata-completeness-heatmap.md`); `proxy.ts` itself is prettier-clean.

### DIFF
--- a/proxy.ts
+++ b/proxy.ts
@@ -8,12 +8,25 @@ import { PUBLIC_PATHS } from "./app/routes/publicPaths";
 // active in production builds despite that empty legacy manifest; don't
 // "verify" middleware presence by reading the legacy file.
 //
-// NOTE: `withAuth` validates the session JWT with `process.env.NEXTAUTH_SECRET`.
-// Unlike the NextAuth API route, it has NO dev fallback secret — if the env var
-// is missing (or differs from what the API route uses), every token check
-// fails and login loops back here forever. The secret is therefore required at
-// runtime in `site-config/hca-atlas-tracker/local/authentication/next-auth-config.ts`,
-// and deployed environments must set NEXTAUTH_SECRET in their task environment.
+// `withAuth` reads the session JWT from the `next-auth.session-token` cookie
+// (`__Secure-`-prefixed under HTTPS) and validates it with
+// `process.env.NEXTAUTH_SECRET`. Unlike the NextAuth API route it has NO dev
+// fallback secret, so NEXTAUTH_SECRET must be set wherever this runs — locally
+// via `site-config/hca-atlas-tracker/local/.env`, and in every deployed task
+// environment.
+//
+// DEBUGGING A LOGIN LOOP (lands back on `/` while `/api/auth/session` shows the
+// user as signed in): the cause is almost never the secret — it's that the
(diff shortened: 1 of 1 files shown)


frano-m (approved):
Nice. Approved.

PR #1462, opened 2026-07-16
fix: white->blue flash on first paint of login/landing page (#1457)

### LINKED ISSUE #1457: fix: white->blue flash on first paint of login/landing page (oversized hero + no background-color fallback)
## Summary
On first view of the login/landing page (`/`, `HomeView`), the page paints **white for up to ~1s, then the blue hero background flashes in**. Two compounding causes: the hero has no solid-color fallback, and the hero image is far larger than it needs to be so its decode is slow.

## Root cause
The blue background is a CSS `background-image` on `StyledSection` (`app/views/HomeView/homeView.styles.ts`):
- **No `background-color` fallback** — until the image decodes and paints, the section is transparent and the white body shows through → the white flash. (`homeView.styles.ts`, `StyledSection`.)
- **Hero image is ~28 MP** — `public/landing/hca-cell.webp` is **7200x3940** (352KB). It's already preloaded (`homeView.tsx` `<link rel="preload" as="image">`), so the *download* isn't the bottleneck — the **decode** of 28 MP is. That delays the first blue paint.

| Version | File | Pixels to decode |
|---|---|---|
| Current 7200x3940 | 352K | 28.4 MP |
| 2560x1400 | 76K | 3.6 MP (~8x less) |
| 1920x1050 | 52K | 2.0 MP (~14x less) |

Image dominant color: **`#004d75`**.

## Possible approaches
These are starting points, not a recommendation — please weigh them and explore other alternatives you think fit better.

**1 — Add a `background-color` fallback.** Set `background-color: #004d75` (the image's dominant color) on `StyledSection` so the very first paint is blue, not white. The image then paints on top of the same color — no visible flash even if decode lags.

**2 — Downsize/optimize the hero image.** 7200x3940 is ~5x wider than any realistic display. Re-export at ~2560px wide (76K, 3.6 MP -> ~8x less decode work) or 1920px (52K, 2.0 MP). Optionally serve responsive variants via `image-set()`/`<picture>` so small screens get a smaller file. This makes the blue paint arrive sooner.

**3 — Low-quality image placeholder (LQIP).** Inline a tiny blurred base64 of the hero as the initial background, swapped for the full image on load. More effort than 1/2.

No preference from Dave on the approach — including combinations or options other than these.

## Notes
- Keep the existing preload; it's correct (`as="image"` works for the CSS background).
- Files: `app/views/HomeView/homeView.styles.ts` (`StyledSection`), `app/views/HomeView/homeView.tsx` (preload), `public/landing/hca-cell.webp`.

frano-m:
## Summary

Closes #1457.

On first view of the login/landing page (`/`, `HomeView`), the page painted **white for up to ~1s, then the blue hero flashed in**. Two compounding causes, both addressed:

### 1. No `background-color` fallback → white flash
The blue hero is a CSS `background-image` on `StyledSection`. Until it decoded, the section was transparent and the white body showed through. Added `background-color: #004d75` (the image's dominant colour) so the **first paint is blue**, and the image paints on top of the same colour — no visible flash even if decode lags.

### 2. Oversized hero → slow decode
`public/landing/hca-cell.webp` was **7200×3940 (358 KB, 28.4 MP)** — ~5× wider than any realistic display, so the decode (not the download) delayed the first blue paint. Re-exported at **2560×1401 (69 KB, 3.6 MP)** — ~8× less decode work and ~81% smaller — so the image paints sooner. The existing preload is kept (correct as-is).

## Verification

Drove the actual landing page with Playwright:
- **Normal load:** section computed `background-color` = `rgb(0, 77, 117)` **and** the hero image is set; renders crisply at the new size.
- **Hero image blocked (route aborted):** section is still `rgb(0, 77, 117)` — blue, not white, with the login card and text fully intact. This deterministically proves the first paint is blue regardless of image decode.

| | before | after |
|---|---|---|
| dimensions | 7200×3940 | 2560×1401 |
| file size | 358 KB | 69 KB |
| pixels to decode | 28.4 MP | 3.6 MP |

- [x] `npm run build:local` succeeds
- [x] Lint / Prettier clean (only pre-existing TODO-tag warnings)
- [x] Playwright verification (normal + image-blocked) confirms blue-first paint


NoopDog:
One follow-up request before merge: **fingerprint the hero image** (content-hashed filename) rather than overwriting it at the same `public/` URL.

**Why it matters here specifically:** this PR replaces `public/landing/hca-cell.webp` in place, so the URL is unchanged. A returning visitor who has the old 352K / 7200×3940 file cached will keep being served those **old cached bytes** — without contacting the server — until their HTTP cache entry expires or revalidates. And since the HTTP cache stores the *encoded* file (not a decoded bitmap), the browser **re-decodes on every load**, so that visitor keeps paying the ~28 MP decode — the exact slow blue paint this PR is fixing — until cache expiry. A content-hashed filename changes the URL, so they download the new 2560×1401 / 68K image immediately and get the fast decode. (The `background-color: #004d75` fallback already helps everyone regardless, since it's CSS.)

**How (Next-native, small change):**
- Move the image out of `public/` into the source tree (e.g. next to `HomeView`).
- `import heroImg from "./hca-cell.webp";`
- Interpolate the URL into the two references:
  ```ts
  background-image: url(${heroImg.src});   // homeView.styles.ts
  href={heroImg.src}                        // preload <link> in homeView.tsx
  ```

Note the one subtlety: a static image import in Next returns a `StaticImageData` object, so use `heroImg.src` (not `heroImg`) in both spots — otherwise you'd get `url([object Object])`. Types work out of the box via `next-env.d.ts`.

Webpack then emits `hca-cell.<contenthash>.webp` and the URL busts automatically on every future image change — no manual version bumps. (A manual rename would also work, but you'd have to remember to bump it each time.)

NoopDog (approved):

PR #1467, opened 2026-07-18
ci: generate next types before doing typechecking but exclude dev types (#1459)

### LINKED ISSUE #1459: Generate Next types before typechecking
Next generates types under `.next/types`, which may cause typechecking (such as is done during the pre-commit hook) to fail or be otherwise inaccurate. `next typegen` can be used to generate just the type definitions. We can add an NPM script that generates Next types and then does typechecking, and use this in place of `npx tsc --noEmit` in places such as the pre-commit hook and the checks GitHub workflow.

Additionally, `.next` should be removed from the ignored TypeScript paths.

Next also creates types under `.next/dev/types`, which are only generated when the dev server is run. It appears, then, to be necessary to exclude the dev types from being checked if we want to avoid having to manually address out-of-sync generated types. This would lead to very slight loss of type safety, because although the dev types and the build types seem to be *almost* identical, the dev types enforce that `_app.tsx` and `_document.tsx` export the correct types of values, while the build types do not. However, since properly generating the dev types to be checked would require running the dev server anyway, and simply loading the app will surface any issues with the app and document modules, the consequences of having unchecked dev types may not warrant concern.

hunterckx:
Closes #1459

I'm not 100% certain I'm happy with excluding the dev types from typechecking (see ticket for context), but I do at least personally prefer that option (with its, AFAIK, minimal consequences) over hard-to-predict commit failures

---

## Summary (Claude-generated)

- Adds a `typecheck` npm script — `next typegen && tsc --noEmit` — so Next's route types are regenerated immediately before every typecheck.
- Replaces the bare `npx tsc --noEmit` with `npm run typecheck` in both the `run-checks` GitHub workflow and the Husky pre-commit hook.
- Narrows the tsconfig `exclude` from `.next` to `.next/dev/types`: the build-time types under `.next/types` (including `validator.ts`, which validates page exports) are now part of the checked program, while the dev-server-only types stay excluded per the tradeoff described in #1459.

### DIFF
--- a/.github/workflows/run-checks.yml
+++ b/.github/workflows/run-checks.yml
@@ -17,7 +17,7 @@ jobs:
           npm ci
           npm run check-format
           npm run lint
-          npx tsc --noEmit
+          npm run typecheck
   test:
     runs-on: ubuntu-latest
     services:
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -18,7 +18,7 @@ npm run lint ||
 )
 
 # Check TypeScript
-npx tsc --noEmit ||
+npm run typecheck ||
 (
 	echo 'TypeScript compile failed';
 	false;
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "build:prod": "./scripts/build.sh hca-atlas-tracker prod && ./scripts/set-version.sh && next build --webpack",
     "start": "next start",
     "lint": "eslint .",
+    "typecheck": "next typegen && tsc --noEmit",
     "check-format": "prettier --check .",
     "prepare": "husky install .husky",
     "test": "jest --runInBand",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -31,5 +31,5 @@
     ".next/types/**/*.ts",
     ".next/dev/types/**/*.ts"
   ],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules", ".next/dev/types"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -31,5 +31,5 @@
     ".next/types/**/*.ts",
     ".next/dev/types/**/*.ts"
   ],
-  "exclude": ["node_modules", ".next"]
+  "exclude": ["node_modules"]
 }


NoopDog:
## Review: generate Next types before typechecking

### Overview

The PR adds a `typecheck` npm script (`next typegen && tsc --noEmit`), swaps it in for the bare `npx tsc --noEmit` in both the CI workflow and the pre-commit hook, and narrows the tsconfig `exclude` from `.next` to `.next/dev/types` so the generated types are actually part of the checked program.

### Verification (ran it locally)

I ran the PR's exact commands with the modified tsconfig:

- `next typegen` succeeds (no env vars needed — `next.config.mjs` is env-free, so CI and pre-commit contexts are fine) and takes only a few seconds, so the hook stays fast.
- `tsc --noEmit` with the new `exclude` **passes cleanly**.
- Comparing `--listFilesOnly` output: the old config only picked up `.next/types/routes.d.ts` (leaked in via the `next-env.d.ts` reference, which bypasses `exclude`). The new config additionally includes `.next/types/validator.ts` and `cache-life.d.ts`. `validator.ts` is the valuable one — it validates every page's exports — so the PR delivers real added type safety, not just plumbing.

### Correctness

The change does what the ticket intends, and the resulting tsconfig matches Next's own recommended setup (Next auto-configures `include: [".next/types/**/*.ts"]` with `.next` not excluded). CI ordering is fine: `check-format` and `lint` run before typegen creates untracked files, and `next-env.d.ts`/`.next` are covered by `.gitignore` and the ESLint ignores.

### Suggestions (all minor)

- **Dead include entry**: `.next/dev/types/**/*.ts` is still in `include` while `.next/dev/types` is in `exclude`. Exclude always wins over include globs, so the include entry is inert and now actively misleading — a reader could think dev types are checked. Suggest dropping it from `include`; the `exclude` entry alone documents the intent.
- **Broadened glob surface** (awareness, not a change request): with `.next` no longer excluded, the root `**/*.ts` include matches *any* `.ts` file under `.next`, not just `.next/types`. Today that's only the three generated type files, but if a future Next version writes stray `.ts` elsewhere under `.next`, typechecking would pick it up. This is inherent to the approach and matches Next's recommended config, so I'd accept it — just flagging.
- **Editor staleness**: the ticket frames staleness as a dev-types problem, but non-dev `.next/types` can also go stale — e.g., delete or rename a page and the editor (which runs plain `tsc` against tsconfig, not the npm script) will error on the stale `validator.ts` until typegen reruns. Worth a line in the README or PR description noting `npm run typecheck` is the fix when the editor shows phantom errors in `.next/types`.
- **Optional hardening**: `next-env.d.ts` is in `.gitignore` and the ESLint ignores but not `.prettierignore`. It currently passes `prettier --check` so nothing breaks, but adding it would insulate `check-format` from future changes to Next's generated header.

### On the stated hesitation about excluding dev types

Excluding dev types looks like the right call. The only safety the dev types add over the build types is validating `_app.tsx`/`_document.tsx` exports, checking them would require a running dev server to stay in sync, and any real breakage there fails instantly on app load. The alternative — unpredictable pre-commit failures depending on whether a dev server ran recently — is strictly worse.

**Verdict**: Approve. The mechanism is sound and verified working; the suggestions above are cleanups, with the dead `include` entry being the one I'd most like to see fixed before merge.

NoopDog (approved):

PR #1497, opened 2026-07-29
refactor: adopt @/ import path alias and migrate deep relative imports (#1495)

### LINKED ISSUE #1495: chore: adopt an enforced import path alias to avoid deep relative imports
## Context

PR #1488's review flagged import-style drift — new files mixed `app/...` (baseUrl) absolute imports with deep relative paths, sometimes in the same file. Those were aligned to the repo's dominant relative convention (~1132 relative vs ~39 `app/` in `app/views`), which fixed the inconsistency but left some very deep chains (up to `../../../../../../../../../../` in the `IntegratedObjectSourceDatasetsView` subtree).

## Proposal

Adopt a single, enforced import-path alias so deep `../../../..` chains go away and the convention is unambiguous:

- Add a `paths` alias in `tsconfig.json` (e.g. `@/*` → `./*`, or formalize the existing `app/*` that already resolves via `baseUrl: "."`).
- Wire it into Jest (`moduleNameMapper`) and any other tooling (Next resolves `tsconfig.paths` natively).
- Add an ESLint rule to enforce it and prevent future drift (e.g. `no-restricted-imports` / a relative-depth limit).
- Codemod existing deep relative imports over to the alias.

## Notes

- `baseUrl: "."` already lets `app/...` resolve, but it's a minority style (~39 usages) and unenforced — hence the drift.
- Decide on **one** convention (alias vs relative) and enforce it repo-wide.

## Related

- Follow-up from the #1488 review (import-drift finding).
- Companion cleanup, now its own ticket: #1496 (repo-wide `as ColumnDef` → `satisfies` swap).

frano-m:
## Summary

Closes #1495

Adopts a repo-root `@/` import path alias and migrates deep relative imports to it, replacing chains like `../../../../../../components/...` with `@/app/components/...`.

## Changes

- **Alias config**
  - `tsconfig.json` — `"@/*": ["./*"]`
  - `jest.config.js` — `moduleNameMapper` `"^@/(.*)$": "<rootDir>/$1"`
  - (Next resolves `tsconfig` paths natively.)
- **Enforcement** — `@typescript-eslint/no-restricted-imports` bans imports 2+ levels up (`^../../`), so deep relative chains can't creep back in. Single-level imports (`./x`, `../x`) stay relative.
- **Codemod** — rewrote **1705 deep imports across 513 files** to the `@/` alias; imports re-sorted by organize-imports (aliases group ahead of `@scope` packages).

## Verification

- `tsc` — 0 errors (all `@/` imports resolve)
- `eslint` — 0 errors (enforcement rule active)
- `prettier` — clean
- `next build` — succeeds
- Full unit test suite — passes (pre-commit hook)

Import paths only — no behavioral change.

## Note

Large mechanical diff (~516 files) — will conflict with other open branches, so worth merging promptly.


NoopDog:
Reviewed with the branch checked out locally — verification first, then one requested change.

**Verified:**

- **The diff is purely mechanical**: outside the three config files, every changed line in all 516 files is an import line (full-diff scan).
- **The codemod respected its own rule exactly**: zero 1-level (`../x`) imports were rewritten. The `app/services/atlases.ts` cases that looked like over-rewrites were pre-existing odd `../../app/...` paths — legitimately 2-level.
- `tsc --noEmit` clean locally; ran a Jest suite that traverses the alias (38/38 pass), so the `moduleNameMapper` works; build/lint/tests green in CI.
- Probed `tsx` directly: it resolves `@/`, so `download-data-dictionary` and tsx-run tooling are fine.

**One real finding — ts-node is alias-blind, and prod runs ts-node:**

Probed concretely: `ts-node` fails with `Cannot find module '@/app/utils/api-errors'` — it doesn't read tsconfig `paths`. Two consequences:

1. **Latent prod landmine.** The container entrypoint runs `npm run migrate` (ts-node) at every startup. Today the migrate chain (`migration-runner.ts` → `pg-migrate-connect-config.ts`; migration files import no app code) contains zero `@/` imports, so nothing breaks now. But the moment a future migration imports app code — which the new ESLint rule will actively steer toward `@/` form — the failure appears **at container startup in prod**, and neither tsc, lint, nor Jest catches it first.
2. **`db_scripts/` are broken under ts-node today** — they import `app/services/*`, which now contain `@/` imports. They work via `tsx`, but their invocation is undocumented.

**Requested change (tiny):** add `tsconfig-paths` as a devDependency and change the migrate scripts to `ts-node -r tsconfig-paths/register ...` (ships to the runtime image via the copied `node_modules`). Makes every ts-node entry point alias-safe and closes both issues at once. Weaker alternative: document "always use tsx" and keep `@/` out of `migrations/**` by convention — but the register flag is one line and removes the class of failure.

Minor: the ESLint regex covers static imports only (dynamic `import("../../x")` isn't caught — currently irrelevant, just noting the boundary). And the "merge promptly" note is right; this should jump the queue ahead of the open dependabot PRs.

Overall: approve once the ts-node path is alias-safe — the mechanical quality of the codemod is excellent.

frano-m:
Great catch on the ts-node prod path — fixed in a39fcaeb.

Added `tsconfig-paths` (devDep — ships to the runtime image via the Dockerfile's `npm ci` + full `node_modules` copy, same as `ts-node`) and `-r tsconfig-paths/register` to both `migrate` and `migrate:down`.

Reproduced and verified your probe:

- **Before:** `ts-node -O '{"module":"commonjs"}' -e "require('@/app/utils/api-errors')"` → `Cannot find module '@/app/utils/api-errors'`
- **After:** same with `-r tsconfig-paths/register` → resolves.

So the container-startup landmine is closed: any future migration importing app code via `@/` now resolves under ts-node.

On the `db_scripts/` note — those run via `tsx`, which (as you confirmed) already resolves `@/`, so they're unaffected by the ts-node gap; only the migrate entrypoint was, and that's now covered.

Noted on the ESLint regex covering static imports only (dynamic `import()` uncaught) — currently irrelevant, will keep in mind if dynamic deep imports ever appear.

NoopDog (approved):

PR #1500, opened 2026-07-30
refactor: use minimal joins between metadata entities and atlases (#1490)

### LINKED ISSUE #1490: Reduce data involved in source dataset grouping
> **Wide jsonb in GROUP BY** — list queries `SELECT d.*` (all jsonb blobs) then `GROUP BY` those rows (`app/data/source-datasets.ts:103,168`, `app/data/component-atlases.ts:101`), pushing `dataset_info`/`validation_info`/etc. through the aggregate. Aggregating in a subquery and joining the blobs back would cut sort/hash memory a lot.

hunterckx:
Closes #1490

### DIFF
--- a/app/data/component-atlases.ts
+++ b/app/data/component-atlases.ts
@@ -100,6 +100,27 @@ export async function getComponentAtlasesForGlobalApi(): Promise<
               PARTITION BY ar.overview->>'network', ar.overview->>'shortName', ar.generation
             ) AS max_revision
           FROM hat.atlases ar
+        ),
+        -- Aggregate atlases by component atlas version separately from the main
+        -- query, so the component atlases' wide jsonb columns don't have to be
+        -- carried through the aggregation.
+        component_atlas_atlases AS (
+          SELECT
+            comp_version_id AS version_id,
+            ARRAY_AGG(
+              jsonb_build_object(
+                'generation', a.generation,
+                'id', a.id,
+                'isPrimary', TRUE, -- TODO: update this when importing is possible
+                'isLatest', a.revision = a.max_revision,
+                'network', a.overview->>'network',
+                'revision', a.revision,
+                'shortName', a.overview->>'shortName'
+              )
+            ) AS atlases
+          FROM atlases_with_revisions a
+          CROSS JOIN UNNEST(a.component_atlases) AS comp_version_id
+          GROUP BY comp_version_id
         )
         SELECT
           ca.*,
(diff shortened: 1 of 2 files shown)


hunterckx:
Re: above two Copilot comments, duplicate IDs in the arrays would be a bug and I'm inclined to think that having duplicate atlases listed for a metadata entity would be a small enough issue that it might actually be better for it to show up so we can notice it

NoopDog (approved):
Thanks @hunterckx — this is a clean implementation of #1490, and it directly validates the approach the ticket proposed. We checked the branch out and reviewed it carefully, including an empirical benchmark. Approving.

## What we verified

**Semantic equivalence (analytical).** The rewrite's correctness hinges on three invariants, and all of them hold:

- `version_id` is the primary key of both `source_datasets` and `component_atlases`. The old `SELECT d.* … GROUP BY d.version_id` already depended on that functional dependency, and the new pre-aggregate-then-join-back-1:1 depends on it identically.
- `atlases.id` is a primary key, so dropping the `GROUP BY` in `getSourceDatasetsForListApi` is safe: with `a.id = $3`, the atlas join contributes at most one row per dataset.
- Switching from `= ANY(array)` joins to `CROSS JOIN UNNEST(array)` would only diverge if an array ever held duplicate version IDs. We checked every write path: atlas arrays only ever get freshly generated UUIDs appended, `ARRAY_REPLACE` swaps values 1:1, removal dedupes via `EXCEPT`, and adding source datasets to an integrated object throws `InvalidOperationError` if any are already present. Duplicates can't occur.

**Result equivalence (empirical).** We seeded a scratch database with 6,000 datasets carrying realistically wide jsonb blobs and 24 atlases (12 families × 2 revisions, ~500-element membership arrays), then ran the old and new global source-datasets queries side by side: 6,000 rows each, symmetric difference zero (with `ARRAY_AGG` order normalized, since it's unspecified in both versions).

**Tests.** All 5 suites exercising the three rewritten queries pass — 97/97.

**Performance.** `EXPLAIN (ANALYZE, BUFFERS)` at that scale with default `work_mem`:

| | Old (main) | New (this PR) |
|---|---|---|
| Execution time | 126 ms | 78 ms |
| Aggregation | GroupAggregate over full wide rows — **32.5 MB external merge sort spilled to disk** | HashAggregate over 48-byte rows (1.4 MB memory, 7.6 MB spill) |
| Shared buffer accesses | 98,798 | 66,075 |

The disk sort of wide jsonb rows — exactly what #1490 predicted — is gone entirely, and the gap should widen as data volume and jsonb width grow.

## Minor observations (non-blocking, no action needed)

- Neither version has an `ORDER BY`, but the old plan happened to emit rows sorted by `version_id` as a side effect of the sort-based aggregate, while the new hash-based plan emits arbitrary (potentially run-to-run-varying) order. Nothing should be relying on the old accidental order — it was uuid order, i.e. meaningless — so this is awareness-only.
- The global CTEs build atlas jsonb objects for every referenced version, including rows the outer `WHERE` later filters out (archived / non-latest). Harmless — the built objects are tiny — and not worth complicating the query over.

Nice work — the pattern of aggregating narrow rows in a CTE and joining the finished arrays back to the PK is exactly the right shape here.

Issue #1504, opened 2026-07-31
chore: upgrade findable-ui to v56.0.0


frano-m:
## Summary

Upgrade `@databiosphere/findable-ui` from `^55.0.0` (currently installed: `55.0.0`) to **v56.0.0**.

Release: https://github.com/DataBiosphere/findable-ui/releases/tag/v56.0.0
Compare: https://github.com/DataBiosphere/findable-ui/compare/v55.0.0...v56.0.0

## ⚠️ Breaking change

- **Error component: "to homepage" button does not return the user to the home page** ([#963](https://github.com/DataBiosphere/findable-ui/issues/963) / [#975](https://github.com/DataBiosphere/findable-ui/pull/975)). Verify our error/500/404 pages and any use of the findable-ui error component still route "to homepage" correctly after the upgrade.

## Other notable changes to verify

- **File rename `entites.ts` → `entities.ts`** ([#985](https://github.com/DataBiosphere/findable-ui/issues/985)). Check we have no deep import referencing the old misspelled path.
- **`@observablehq/plot` marked optional in `peerDependenciesMeta`** ([#988](https://github.com/DataBiosphere/findable-ui/issues/988)) and **ChartView code-split so plot isn't bundled for chart-less consumers** ([#989](https://github.com/DataBiosphere/findable-ui/issues/989)). Confirm any ChartView usage still resolves `@observablehq/plot` (add it as a direct dep if we use charts).
- Perf: memoized virtualized table rows ([#992](https://github.com/DataBiosphere/findable-ui/issues/992)); grouped mobile card re-render fix on sub-row selection ([#995](https://github.com/DataBiosphere/findable-ui/issues/995)) — sanity-check our tables/mobile cards.
- Chore: non-breaking dependency + prettier upgrades.

## Tasks

- [ ] `npm install @databiosphere/findable-ui@^56.0.0` and update `package.json` / lockfile
- [ ] Resolve any type/import breakages (esp. the error component and the `entities.ts` rename)
- [ ] Verify error pages' "to homepage" behavior against the breaking change
- [ ] Verify ChartView / `@observablehq/plot` resolution if charts are used
- [ ] Smoke-test tables, mobile grouped cards, and auth flows (findable-ui auth components)
- [ ] Build (`npm run build:local`), `npm run lint`, `npm run check-format`, `npm test`, `npx playwright test`

## Notes

[...]

Issue #1477, opened 2026-07-23
s3-sync.test.ts fails on non-UTC machines — JS clock compared against zoneless DB timestamps


NoopDog:
## Problem

The test `syncFilesFromS3 › processes s3 data, logs warnings, and creates and validates files as appropriate` (`__tests__/s3-sync.test.ts`) fails consistently on developer machines in non-UTC timezones, while passing in CI:

```
expect(files).toHaveLength(8);   // __tests__/s3-sync.test.ts:453
Expected length: 8
Received length: 0
```

Reproduced on current `main` on a UTC+10 machine; the same suite passes on UTC runners.

## Root cause

The test captures a start time in JavaScript and compares it against Postgres-written timestamps:

- `const startTime = new Date();` (`__tests__/s3-sync.test.ts:410`)
- `SELECT * FROM hat.files WHERE updated_at >= $1` with that Date (`:588-596`)

But `hat.files.updated_at` is `timestamp without time zone`, populated by `CURRENT_TIMESTAMP` (schema.sql:54, and the `update_updated_at_column()` trigger). Casting the parameterized JS Date to a zoneless timestamp drops its offset, so the comparison is between naive wall-clock values from two different clocks (Node's local time vs. the Postgres session timezone). With a UTC Postgres and a UTC+10 machine, rows are stamped ~10 hours "earlier" than `startTime`, and the `updated_at >= $1` filter matches nothing. The later assertion at `:487` (`fileTime > startTime.getTime()`) has the same latent problem.

## Proposed fix

Take `startTime` from the database instead of the JS clock — e.g. `SELECT CURRENT_TIMESTAMP` at the start of the test — so both sides of the comparison come from one clock. That's a test-only change.

Longer-term (separate decision): the schema uses `timestamp without time zone` throughout; migrating to `timestamptz` would remove this class of bug from app code too, but that's a larger change and not needed to fix the test.

## Impact

Any developer outside UTC gets a red `npm test` on a clean checkout of `main`, which masks real failures and erodes trust in the suite.

hunterckx:
Closing as duplicate of #1464 (which I will note seems to more accurately describe the problem as coming from the difference in timezone between Node and Postgres -- the test runs fine for me in Pacific Time)

PR #1509, opened 2026-07-31
chore: drop @observablehq/plot dependency (#1470)

### LINKED ISSUE #1470: `@observablehq/plot` required only to satisfy findable-ui peer deps — mark optional upstream
## Problem

`@observablehq/plot` is a direct dependency of the tracker but is never imported by tracker code. It exists solely to satisfy `@databiosphere/findable-ui@55.0.0`'s `peerDependencies`, which declares it as **required** — only `next-auth` is marked optional in `peerDependenciesMeta`.

Within findable-ui, `@observablehq/plot` is used only by the `ChartView`/`Plot` components (`components/Plot/components/BarX`, `components/Index/.../ChartView`), which the tracker never uses (no `ChartView`/`BarX` references in `app/` or `site-config/`). We install it purely to satisfy the peer declaration.

## Proposed fix

1. **Upstream (findable-ui):** mark `@observablehq/plot` as optional in `peerDependenciesMeta` (as `next-auth` already is), since it's only needed by consumers that use `ChartView`/`Plot`. Other feature-specific peers may deserve the same treatment.
2. **Tracker (after upstream release):** drop `@observablehq/plot` from `dependencies`.
3. **Verify (cheap, can do now):** run a bundle analyzer to confirm `@observablehq/plot` is tree-shaken out of the client bundle today — if it isn't, this is a real bundle-size win, not just dependency hygiene.

## Context

Found during a performance review (see #1458 for the database half).

frano-m:
## Summary

Removes `@observablehq/plot` as a tracker dependency. The tracker never renders findable-ui's `ChartView`/`Plot`, and **v56 made `@observablehq/plot` an optional peer** — so npm no longer installs it and it can be dropped.

Closes #1470

## Why this now works (it didn't before v56)

- In findable-ui **v55**, `@observablehq/plot` was a **required** peer, so npm installed it everywhere. #1470's upstream half (mark it optional) shipped in **v56** (already on main via #1504).
- npm does **not** auto-install *optional* peers. Confirmed two ways:
  - A clean install here with plot removed from `package.json` leaves it out entirely (`node_modules`, lockfile).
  - brc-analytics (another findable-ui consumer that doesn't use `next-auth`, findable-ui's *other* optional peer) has **no `next-auth` in `node_modules`** — same mechanism.

## Changes

- **`package.json`** — remove `@observablehq/plot`.
- **`package-lock.json`** — drops the whole d3/plot subtree (~490 lines), no other churn.
- **`next.config.mjs`** — when `@observablehq/plot` is **not installed**, alias it to `false` (empty module). findable-ui's `ExploreView` still *statically compiles* its lazy `ChartView` chunk at build time, so webpack must resolve the module even though it's never bundled for our chart-less pages. The alias lets the build succeed with the package absent. The stub is applied **conditionally** (probed via `require.resolve`): a future `npm install @observablehq/plot` restores the real module with no further config change — an unconditional stub would silently shadow the installed package.

## Verification

- `npm run build:local` compiles successfully with plot absent (clean install, plot not in `node_modules`); build also succeeds when plot *is* installed (alias skipped).
- `tsc`, ESLint, Prettier clean; **1196/1196** unit tests pass.
- `ExploreView` renders correctly (manual check) — the lazy chart chunk (which the alias stubs) is only loaded when a user switches to the chart view, which the tracker never exposes.

## Notes

- **Future ChartView adoption:** if a page ever renders findable-ui's `ChartView`, install `@observablehq/plot` — the conditional alias then steps aside automatically. Rendering a chart *without* installing it would resolve `Plot.plot` to `undefined` and **throw a `TypeError` at render** (dropping the view to the error boundary), not render a blank chart. Documented in the `next.config.mjs` comment.
- **Turbopack:** the alias lives in the `webpack` config, which Next 16 honors only under `--webpack`. Every npm script forces `--webpack`, so it always applies; a bare `next dev` (Turbopack, the Next 16 default) would ignore it. Turbopack's `resolveAlias` has no `false` equivalent, so mirroring it would need a maintained stub file — avoided; the coupling is documented in the config comment instead.
- **Durable upstream fix** tracked in DataBiosphere/findable-ui (ExploreView "view slot" so chart-less consumers never compile the chunk) — see linked spike.

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -39,4 +39,21 @@ export default withMDX({
     ];
   },
   transpilePackages: [...ESM_PACKAGES],
+  /**
+   * Extend the webpack configuration.
+   * @param config - Webpack configuration.
+   * @returns The modified webpack configuration.
+   */
+  webpack: (config) => {
+    // @observablehq/plot is only reached through findable-ui's lazy ChartView
+    // chunk, which the tracker never renders (no ChartView/Plot usage in app/ or
+    // site-config/). The package is not installed; alias it to an empty module
+    // so webpack can still compile the lazy chunk at build time instead of
+    // failing to resolve it. See clevercanary/hca-atlas-tracker#1470.
+    config.resolve.alias = {
+      ...config.resolve.alias,
+      "@observablehq/plot": false,
+    };
+    return config;
+  },
 });
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -1,4 +1,27 @@
 import nextMDX from "@next/mdx";
+import { createRequire } from "node:module";
+
+const require = createRequire(import.meta.url);
+
+// Detect whether @observablehq/plot is installed. It is an optional peer of
+// findable-ui that the tracker never renders (no ChartView/Plot usage in app/
+// or site-config/); it was dropped as a direct dependency. findable-ui still
+// statically references it from its lazy ChartView chunk, so the bundler must
+// resolve the specifier at build time even though that chunk is never
+// downloaded. When plot is absent we alias it to an empty module (below) so the
+// build succeeds. The alias is applied ONLY when plot is not installed: a future
+// consumer who runs `npm install @observablehq/plot` to actually render charts
+// would otherwise hit an unconditional stub silently shadowing the real package
+// (`Plot.plot` resolves to undefined and throws a TypeError at render, dropping
+// the view to the error boundary with no build-time signal). Probing collapses
+// that future fix to the single obvious step of installing the package.
+let plotInstalled;
+try {
+  require.resolve("@observablehq/plot");
+  plotInstalled = true;
+} catch {
+  plotInstalled = false;
+}
 
 const ESM_PACKAGES = [
   "codsen-utils",
@@ -39,21 +62,26 @@ export default withMDX({
     ];
   },
   transpilePackages: [...ESM_PACKAGES],
-  /**
-   * Extend the webpack configuration.
-   * @param config - Webpack configuration.
-   * @returns The modified webpack configuration.
-   */
   webpack: (config) => {
-    // @observablehq/plot is only reached through findable-ui's lazy ChartView
-    // chunk, which the tracker never renders (no ChartView/Plot usage in app/ or
-    // site-config/). The package is not installed; alias it to an empty module
-    // so webpack can still compile the lazy chunk at build time instead of
-    // failing to resolve it. See clevercanary/hca-atlas-tracker#1470.
-    config.resolve.alias = {
-      ...config.resolve.alias,
-      "@observablehq/plot": false,
-    };
+    // Stub the absent @observablehq/plot so webpack can compile findable-ui's
+    // lazy ChartView chunk (see the plotInstalled probe above for the full
+    // rationale). Only applied when plot is not installed, so installing it
+    // later restores the real module without further config changes.
+    //
+    // This lives in the webpack config, which Next 16 runs only under
+    // `--webpack`. The scripts that compile — `dev` and `build:*` — all force
+    // `--webpack`, so it always applies where bundling happens (`start` just
+    // serves prebuilt output, no bundling). A bare `next dev` uses Turbopack
+    // (the Next 16 default), which ignores this function and would fail to
+    // resolve the ChartView chunk. Turbopack's resolveAlias has no `false`
+    // equivalent, so mirroring it would require a maintained empty-stub file —
+    // deliberately avoided here.
+    if (!plotInstalled) {
+      config.resolve.alias = {
+        ...config.resolve.alias,
+        "@observablehq/plot": false,
+      };
+    }
     return config;
   },
 });
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,6 @@
         "@mui/material": "^7",
         "@mui/material-nextjs": "^9",
         "@next/mdx": "^16",
-        "@observablehq/plot": "^0",
         "@tanstack/match-sorter-utils": "^8",
         "@tanstack/react-query": "^5.101.4",
         "@tanstack/react-table": "^8",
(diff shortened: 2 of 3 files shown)


NoopDog:
Thanks @frano-m — clean, minimal completion of #1470's consumer half, and the PR description's verification claims all held up under independent re-verification. We ran the full review suite on a local checkout (multi-agent simplification review, security review, code review, plus a fresh `npm ci`/build/test gate). The change is correct as-is; the asks below all target one theme — the alias works perfectly today but leaves a few latent traps for future developers, each cheap to defuse now.

## What checked out 👍

- **Security: no findings.** The two lockfile version refreshes are ordinary in-range patch bumps resolving to registry.npmjs.org with intact sha512 integrity hashes; the stubbed path is unreachable at runtime (no `enableEntitiesView` anywhere in `site-config/`); the alias affects no other module resolution. Net effect is a *smaller* installed attack surface.
- **`alias: false` is the right mechanism** — strictly better than `IgnorePlugin` (throws at require time) or a replacement plugin (needs a maintained stub file). No upstream helper or consumer-repo precedent exists; this is the first findable-ui consumer to drop the package.
- **Measured payoff:** 41 packages (~2.5 MB unpacked) removed from every `npm ci`; zero plot/d3 bytes anywhere in `.next/` output; the residual stub chunk is 5.2 KB, emitted but never downloaded (referenced only by the lazy import that never fires).
- **`react-window` correctly left in place** — verified it's a *required* peer (`peerDependenciesMeta` marks only plot and next-auth optional) genuinely imported by findable-ui Filter components the tracker renders. Not the same class as plot.
- **Jest needs no matching stub** — `next/jest` doesn't apply webpack aliases, but the lazy chunk only resolves if a test renders ChartView, and none does. (If one ever did, the existing `moduleNameMapper` + `__mocks__/` mechanism is the right home.)
- **Independent gate:** clean install with plot absent from `node_modules`; `npm run build:local` ✅; tests 1195/1196 — the sole real failure is the pre-existing #1477 timezone issue (plus one load flake that passed in isolation; this PR touches no runtime TS).

## Asks (small; 1–2 feel worth doing in this PR)

1. **Correct the documented failure mode.** The PR body and the `next.config.mjs` comment say a future ChartView adoption "would otherwise render an empty chart" — but tracing the code, the stub actually makes `Plot.axisX`/`Plot.plot` throw a render-time `TypeError` deep in findable-ui internals, crashing the explore view to the error boundary with no build-time signal. Two independent review passes converged on this. The comment is the future dev's only map from symptom to cause, so it should describe the real symptom.
2. **Make the alias conditional (~5 lines).** Probe with `createRequire(import.meta.url)` + `require.resolve("@observablehq/plot")` and only apply the stub when the package is absent. This defuses the nastiest trap: today the natural fix — `npm install @observablehq/plot` — does *not* work, because the alias stubs the module even when it's installed. Conditional aliasing collapses the future fix from two coupled steps ("remove alias AND reinstall") to one obvious one.
3. **Cover the Turbopack gap.** Next 16 defaults to Turbopack, which silently ignores the `webpack` function. Every script currently forces `--webpack`, so this is latent — but one bare `next dev` fails with "Module not found" in the ChartView chunk. Adding the equivalent `turbopack.resolveAlias` entry alongside (or at minimum a comment documenting the `--webpack` coupling) closes it.
4. **File the upstream findable-ui issue for the durable fix** — the ExploreView "view slot" (dependency inversion so chart-less consumers never compile the chunk), named in #1470's comments but not tracked anywhere open. This PR closes #1470, after which the only pointer to the durable fix lives in a closed issue and the shim becomes permanent by default. Referencing the upstream issue from the `next.config.mjs` comment keeps the trail alive.
5. Trivial: the JSDoc block on `webpack` only restates the signature — `next.config.mjs` is in ESLint's ignore list so it isn't lint-mandated, and the inline comment below it carries all the substance. Fine to drop.

None of these block the change's correctness today — the stubbed path is unreachable and the gate is green. Nice work landing the measurable half of #1470.

frano-m:
Thanks for the thorough re-verification, @NoopDog — all five addressed in `ad3b008e`:

1. **Failure mode corrected.** The `next.config.mjs` comment and the PR body now describe the real symptom: with the stub in place, `Plot.plot` resolves to `undefined` and **throws a `TypeError` at render** (dropping the view to the error boundary), rather than rendering a blank chart. Good catch — the old wording would have sent a future dev looking in the wrong place.

2. **Alias is now conditional.** Probe with `createRequire(import.meta.url)` + `require.resolve("@observablehq/plot")`; the `false` stub is applied only when the package is absent. So a future `npm install @observablehq/plot` restores the real module with no config change — the fix collapses to one obvious step.

3. **Turbopack coupling documented (comment, not a `resolveAlias` mirror).** Turbopack's `resolveAlias` has no `false` equivalent, so a real mirror would need a maintained empty-stub file — which is exactly the "maintained stub file" tradeoff you noted `alias: false` avoids. Since every npm script forces `--webpack`, I documented the coupling in the config comment (your explicit acceptable-minimum) rather than reintroduce a stub for the unreachable Turbopack path. Happy to add the stub + `turbopack.resolveAlias` if you'd rather close it fully.

4. **Upstream spike filed:** DataBiosphere/findable-ui#997 — the ExploreView "view slot" (dependency inversion so chart-less consumers never compile the chunk). Referenced from the PR body so the trail survives #1470 closing.

5. **Redundant JSDoc dropped** from the `webpack` function; the inline comment carries the substance.

Also removed the ticket-number references from the config comments per a follow-up. Ready for another look.

frano-m:
Copilot's suppressed comment on `next.config.mjs` was right: `npm start` runs `next start` (serves prebuilt output, no bundling), so it doesn't take `--webpack` and shouldn't have been in that list. Corrected the comment to scope the `--webpack` coupling to the scripts that actually compile — `dev` and `build:*` — and note that `start` just serves the built output. Pushed in `0e5d2a89`.

NoopDog (approved):
Thanks @frano-m — all five asks addressed, and each one re-verified on a fresh checkout. Approving.

1. **Failure mode corrected** ✅ — the config comment and PR body now describe the real symptom (`Plot.plot` resolves to `undefined` → render-time `TypeError` → error boundary), not a blank chart. The trail from symptom to cause is now honest.

2. **Conditional alias verified empirically, in both directions.**
   - Plot absent (`npm ci` on this branch): `build:local` succeeds; the lazy chart chunk is the 5 KB alias stub.
   - Plot installed (`npm install @observablehq/plot --no-save`, cold cache): `build:local` succeeds; the stub chunk is gone, replaced by a real ~245 KB chunk with plot/d3 bundled.

   So `npm install @observablehq/plot` alone genuinely restores charts with zero config changes — exactly the collapse-to-one-step behavior the review asked for. (One testing footnote, not a defect: a warm webpack cache briefly kept serving the stale stub after install in our probe, but only because `--no-save` skipped the lockfile write; a real install updates the lockfile, which invalidates Next's build cache, so the realistic flow is unaffected.)

3. **Turbopack coupling documented rather than mirrored — agreed with the reasoning.** Turbopack's `resolveAlias` has no `false` equivalent, so a true mirror would require exactly the maintained-stub-file tradeoff that `alias: false` exists to avoid. Every compiling script forces `--webpack`, the comment documents the coupling (correctly scoped to `dev`/`build:*` after the Copilot catch about `npm start`), and that was the acceptable minimum named in the review. No need for the stub.

4. **Upstream spike confirmed**: DataBiosphere/findable-ui#997 (ExploreView "view slot" so chart-less consumers never compile the chunk) — open, and referenced from the PR body so the durable-fix trail survives #1470 closing. Moving the ticket references out of the code comments and into the PR body was the right call per convention; the comment stands on its own.

5. **Boilerplate JSDoc dropped** ✅.

Nice tight turnaround — the conditional probe in particular turned the config from a hidden trap into something self-healing.

Issue #1507, opened 2026-07-31
chore: enforce @/ alias import convention — ban all ../ relative imports


frano-m:
## Summary

Tighten the intra-repo import convention so it stops being a review conversation: **relative (`./`) only for same-directory and descendants; `@/` alias for everything else — no `../` ever.**

## Current state

`eslint.config.mjs` already has `@typescript-eslint/no-restricted-imports` banning **deep** relatives only:

```js
{ regex: "^\\.\\./\\.\\./", message: "Use the '@/' path alias instead of deep relative imports (2+ levels up)." }
```

So `../../foo` is banned but single-level `../foo` is still allowed. `@/*` is configured (`tsconfig.json` paths), and **514 files** already use it; **124 files** still use single-level `../`.

## Proposed change

Tighten the regex from `^\.\./\.\./` (2+ levels) to `^\.\./` (any parent) so **all** `../` imports are banned — relative stays legal only for `./` (same dir) and its descendants.

Rationale (from brc-analytics#1573): the rule describes the *relationship* — am I reaching outside my own subtree? — not incidental file depth, so it doesn't change when files move; and an internal reference to a shared module is the same string as an external one, so a single `grep` finds every consumer.

## Tasks

- [ ] Update the `no-restricted-imports` pattern to `^\.\./` (ban all parent-relative imports).
- [ ] Sweep the ~124 files still using `../` → `@/…` alias.
- [ ] Check in a short version of the convention alongside the lint config (or in `CODING_STANDARDS.md`).
- [ ] Confirm `prettier` / `prettier-plugin-organize-imports` play nicely with the resulting ordering.

Mirrors [galaxyproject/brc-analytics#1573](https://github.com/galaxyproject/brc-analytics/issues/1573). Pairs with #1506 (type-only imports).

PR #1524, opened 2026-08-05
chore: drop tsconfig baseurl so bare repo-root imports fail to resolve (#1511)

### LINKED ISSUE #1511: chore: drop tsconfig baseUrl to make bare repo-root imports fail at resolution (retire the lint regex)
## Summary

Follow-up to #1507. That PR banned bare repo-root imports (e.g. `import { x } from "app/foo"`, resolved via `baseUrl: "."`) with a `no-restricted-imports` **regex** that enumerates the top-level source dirs. The regex works and gives an actionable message, but it's a denylist that must be kept in sync with the repo's directories by hand.

The deeper fix removes the mechanism the bare imports rely on, so they can't resolve at all:

- **Drop `baseUrl` from `tsconfig.json`.** TypeScript ≥ 4.1 resolves `paths` (`@/*`) without `baseUrl`, so the `@/` alias keeps working; bare `app/foo` stops resolving and fails to compile.
- **Trim `<rootDir>/` from Jest `moduleDirectories`** in `jest.config.js` (currently `["node_modules", "<rootDir>/"]`), which is what independently lets bare specifiers resolve in tests.

With both gone, the `no-restricted-imports` bare-root regex added in #1507 becomes unnecessary and can be removed (the `..`/`../**` parent-relative pattern stays).

## Trade-off

The lint rule gives a pointed message ("Import repo-root modules via the '@/' alias"); dropping `baseUrl` instead surfaces TypeScript's generic "Cannot find module 'app/foo'". Acceptable — the failure is at compile/type time and the fix (`@/app/foo`) is obvious — but noting it so the decision is deliberate.

## Tasks

- [ ] Remove `baseUrl` from `tsconfig.json`; confirm `@/*` still resolves (build, typecheck).
- [ ] Remove `<rootDir>/` from `moduleDirectories` in `jest.config.js`; confirm tests resolve `@/` and no bare-root imports remain.
- [ ] Remove the bare-root regex from `no-restricted-imports` in `eslint.config.mjs`.
- [ ] Full gate: build, typecheck, lint, tests.

Raised in review of #1507 (PR #1510) by @NoopDog; two review agents converged on it independently.

frano-m:
## Summary

Closes #1511. Follow-up to #1507.

#1507 banned bare repo-root imports (e.g. `import { x } from "app/foo"`, resolved via `baseUrl: "."`) with a `no-restricted-imports` **regex** enumerating the top-level source dirs Ã¢ÂÂ a hand-maintained denylist. This removes the *mechanism* those imports rely on, so they can't resolve at all, and retires the regex.

## Changes

- **`tsconfig.json`** Ã¢ÂÂ removed `"baseUrl": "."`. TypeScript resolves `paths` (`@/*`) relative to the tsconfig dir without it, so the alias keeps working; a bare `app/foo` now fails to compile with `TS2307: Cannot find module 'app/foo'`.
- **`jest.config.js`** Ã¢ÂÂ `moduleDirectories: ["node_modules"]` (dropped `<rootDir>/`), which independently let bare specifiers resolve in tests. `@/` still resolves via `moduleNameMapper` (`^@/(.*)$` Ã¢ÂÂ `<rootDir>/$1`), unaffected.
- **`eslint.config.mjs`** Ã¢ÂÂ removed the bare-root regex from `no-restricted-imports` (the `..`/`../**` parent-relative pattern stays), replaced with a short comment noting bare-root now fails to resolve.

## Trade-off

The lint rule gave a pointed message ("Import repo-root modules via the '@/' alias"); dropping `baseUrl` instead surfaces TypeScript's generic `Cannot find module 'app/foo'`. Acceptable Ã¢ÂÂ it fails at compile/type time and the fix (`@/app/foo`) is obvious Ã¢ÂÂ and it's not a denylist that rots as directories are added.

## Verification

`@/` still resolves everywhere; a bare-root import now fails:

| Path | Result |
|---|---|
| `typecheck` | `@/` resolves; bare `app/...` Ã¢ÂÂ `TS2307` (verified with a probe file) |
| `build:local` | succeeds (next/webpack) |
| `test` | 1211/1211 (jest `@/` via `moduleNameMapper`; note: jest is NOT a backstop for bare imports â see below) |
| `ts-node -r tsconfig-paths/register` (migrate) | `@/` resolves (tsconfig-paths defaults baseUrl to the config dir) |
| `tsx` (db scripts) | `@/` resolves |
| `lint` / `check-format` | 0 errors / clean |

**Enforcement is typecheck-only (which is sufficient).** A static bare import still resolves in jest — with no `baseUrl`, Next computes an implicit baseUrl (the tsconfig dir) and its SWC transform rewrites the specifier before jest's resolver runs — so jest is not a backstop. The `moduleDirectories` change only closes runtime `require`/`jest.mock` bare paths. The real guard is `typecheck` (TS2307), run by CI (`run-checks.yml`) and the pre-commit hook.

No new tests: the change is config, and CI's `typecheck` Ã¢ÂÂ which now rejects bare-root imports Ã¢ÂÂ is itself the durable guard (a jest test can't cleanly assert a compile failure).

Ã°ÂÂ¤Â Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -73,10 +73,13 @@ const config = [
               message:
                 "Reach outside this directory via the '@/' alias; relative imports are for ./ same-dir and descendants only.",
             },
-            // Bare repo-root imports (e.g. `app/foo`) need no lint rule: tsconfig
-            // has no `baseUrl` and jest's `moduleDirectories` omits `<rootDir>`,
-            // so they fail to resolve at compile/test time. Only the `@/` alias
-            // reaches repo-root modules.
+            // Bare repo-root imports (e.g. `app/foo`) need no lint rule: with no
+            // `baseUrl` in tsconfig they fail typecheck (TS2307), which CI and
+            // the pre-commit hook enforce. (jest's SWC transform still resolves a
+            // static bare import via Next's implicit baseUrl, so jest is not a
+            // backstop; the trimmed `moduleDirectories` only closes runtime
+            // `require`/`jest.mock` paths.) Only the `@/` alias reaches repo-root
+            // modules.
           ],
         },
       ],
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -73,17 +73,10 @@ const config = [
               message:
                 "Reach outside this directory via the '@/' alias; relative imports are for ./ same-dir and descendants only.",
             },
-            {
-              // Bare repo-root imports resolve via `baseUrl: "."` (e.g.
-              // `app/foo`), which is a second specifier for the same module —
-              // defeating the one-string/one-grep goal above. Require the `@/`
-              // alias for every repo-root segment. Keep this list in sync with
-              // the repo's top-level source directories.
-              message:
-                "Import repo-root modules via the '@/' alias (e.g. '@/app/...'), not a bare baseUrl path.",
-              regex:
-                "^(@types|__mocks__|__tests__|app|catalog|db_scripts|migrations|pages|scripts|site-config|testing|types)/",
-            },
+            // Bare repo-root imports (e.g. `app/foo`) need no lint rule: tsconfig
+            // has no `baseUrl` and jest's `moduleDirectories` omits `<rootDir>`,
+            // so they fail to resolve at compile/test time. Only the `@/` alias
+            // reaches repo-root modules.
           ],
         },
       ],
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,7 +12,9 @@ const createJestConfig = nextJest({
 const customJestConfig = {
   // Add more setup options before each test is run
   // setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
-  // if using TypeScript with a baseUrl set to the root directory then you need the below for alias' to work
+  // node_modules only (no <rootDir>): closes runtime resolution of bare
+  // repo-root specifiers (e.g. require("app/foo")) now that tsconfig has no
+  // baseUrl. The `@/` alias resolves via moduleNameMapper below regardless.
   moduleDirectories: ["node_modules"],
   moduleNameMapper: {
     "^@/(.*)$": "<rootDir>/$1",
--- a/jest.config.js
+++ b/jest.config.js
@@ -13,7 +13,7 @@ const customJestConfig = {
   // Add more setup options before each test is run
   // setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
   // if using TypeScript with a baseUrl set to the root directory then you need the below for alias' to work
-  moduleDirectories: ["node_modules", "<rootDir>/"],
+  moduleDirectories: ["node_modules"],
   moduleNameMapper: {
     "^@/(.*)$": "<rootDir>/$1",
     "^ky$": "<rootDir>/__mocks__/ky.ts",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,7 +14,6 @@
     "isolatedModules": true,
     "jsx": "react-jsx",
     "incremental": true,
-    "baseUrl": ".",
     "paths": {
       "@/*": ["./*"]
     },


NoopDog (approved):
Thanks @frano-m — this is the right durable fix from #1511, and the risky parts all verified clean on a local checkout. Approving, with one accuracy correction to make in the PR body and the `eslint.config.mjs` comment before merge (text-only, no code change needed).

## Verified

- **The riskiest path holds**: `npm run migrate` (`ts-node -r tsconfig-paths/register` — the prod container's entrypoint) resolves `@/` without `baseUrl`, and so do `tsx`, `tsc`, the webpack build, and jest. Full gate: **1211/1211 tests**, `tsc` clean, `build:local` ✅, ESLint 0 errors, Prettier clean.
- **The goal is achieved**: a bare `app/...` / `testing/...` import now fails typecheck with `TS2307` — probe-verified in both `app/` and `__tests__/` — and since CI (`run-checks.yml`) and the husky pre-commit hook both run `typecheck`, nothing bare can merge. The hand-maintained regex denylist is correctly retired.

## The one ask: correct the "fails at test time" claim

We probed a static `import { TEST_USERS } from "testing/constants"` inside a jest test on this branch — **it still resolves and the test passes**, despite `moduleDirectories: ["node_modules"]`. The mechanism: when `baseUrl` is absent, Next computes an *implicit* baseUrl (the tsconfig's directory — see `resolvedBaseUrl` / `isImplicit: true` in `next/dist/build/load-jsconfig.js`) and feeds it to the SWC transform that `next/jest` uses, which rewrites bare specifiers at transform time — before jest's resolver ever runs. Only **runtime** resolution now fails (`require.resolve("testing/constants")` in a test throws, so `require`/`jest.mock`-style paths are genuinely closed).

Net effect:
- Enforcement for static bare imports is **typecheck-only** — which is sufficient (CI + pre-commit), so the change stands.
- But the `eslint.config.mjs` comment ("jest's `moduleDirectories` omits `<rootDir>`, so they fail to resolve at compile/test time") and the PR-body verification table overstate the jest half. Please reword to something like: *"bare repo-root imports fail typecheck (TS2307), which CI and pre-commit enforce; jest's SWC transform would still resolve a static bare import via Next's implicit baseUrl, but runtime `require`/`jest.mock` paths are closed by the `moduleDirectories` change."* A future reader shouldn't rely on jest as a backstop that isn't there.

The `moduleDirectories` change itself is worth keeping exactly as-is — it closes the runtime half.

## Nit (take or leave)

`jest.config.js` still carries the boilerplate comment "if using TypeScript with a baseUrl set to the root directory then you need the below for alias' to work" above `moduleNameMapper` — now stale, since there is no baseUrl and the mapper serves the `@/` alias regardless.

Nice completion of the #1507 → #1510 → #1511 arc — the convention is now enforced by the resolver rather than a list someone has to remember to extend.

frano-m:
Thanks @NoopDog — and good catch, you're exactly right. Corrected in `7b2c5c35` (text-only).

I re-probed to confirm before rewording:

```
STATIC_BARE:  RESOLVED (19)      // import { TEST_USERS } from "testing/constants" in a jest test
RUNTIME_BARE: THREW (closed)     // require.resolve("testing/constants")
```

So a static bare import still resolves in jest via Next's implicit baseUrl + SWC transform (before jest's resolver runs) — jest is **not** a backstop. The `moduleDirectories` trim only closes runtime `require`/`jest.mock` paths (worth keeping, as you said). The real, sufficient enforcement is **typecheck** (TS2307), run by CI (`run-checks.yml`) + pre-commit.

- Reworded the `eslint.config.mjs` comment to say bare imports fail **typecheck** (not "compile/test time"), and that jest's SWC transform still resolves static bare imports so it isn't a backstop.
- Updated the PR body verification section with the same clarification.
- Nit: replaced the stale "if using TypeScript with a baseUrl…" boilerplate above `moduleNameMapper` in `jest.config.js` — it now describes what `moduleDirectories: ["node_modules"]` actually does (closes runtime bare paths) and that `@/` resolves via the mapper regardless of baseUrl.

No code change; gate still green (1211/1211, `tsc` clean, ESLint 0 errors, Prettier clean).

